PR #8, opened 2015-02-10
git: fixing export when a repo has multiple remotes

jbohren:
Currently when running `vcstool export`, if a git repo has multiple remotes, the `git remote show` command returns a list of remotes and `vcstool` fails to determine which remote should be used in the export.

This instead parses the output of `git branch -vv` which gives the following information:

* The currently checked-out branch
* The name of the remote of that branch
* The name of the remote branch of that local branch

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -1,5 +1,6 @@
 import copy
 import os
+import re
 
 from .vcs_base import find_executable, VcsClientBase
 
@@ -64,12 +65,23 @@ class GitClient(VcsClientBase):
         }
 
     def _get_url(self):
-        cmd_remote = [GitClient._executable, 'remote', 'show']
+        cmd_remote = [GitClient._executable, 'branch', '-vv', '--color=never']
         result_remote = self._run_command(cmd_remote)
         if result_remote['returncode']:
             result_remote['output'] = 'Could not determine remote: %s' % result_remote['output']
             return result_remote
-        remote = result_remote['output']
+        branches = result_remote['output']
+
+        for line in branches.splitlines():
+            tokens = line.split()
+            if len(tokens) > 3 and tokens[0] == '*':
+                ref = re.findall('\[(.+)\]',tokens[3])
+                if len(ref) == 1:
+                    remote, remote_branch = ref[0].split('/',1)
+                else:
+                    result_remote['output'] = 'Could not determine remote: %s' % result_remote['output']
+                    return result_remote
+                break
 
         cmd_url_tpl = [GitClient._executable, 'config', '--get', 'remote.%s.url']
         cmd_url = copy.copy(cmd_url_tpl)
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -34,68 +34,137 @@ class GitClient(VcsClientBase):
         return self._run_command(cmd)
 
     def export(self, command):
-        result_url = self._get_url()
+
+        # Get the remote and remote reference
+        result_remote = self._get_remote_ref(command.exact)
+        if result_remote['returncode']:
+            return result_remote
+
+        remote = result_remote['output'][0]
+        ref = result_remote['output'][1]
+
+        # Get the url for this remote
+        result_url = self._get_url(remote)
         if result_url['returncode']:
             return result_url
+
         url = result_url['output'][0]
 
-        cmd_ref = [GitClient._executable, 'rev-parse', 'HEAD']
-        result_ref = self._run_command(cmd_ref)
-        if result_ref['returncode']:
-            result_ref['output'] = 'Could not determine ref: %s' % result_ref['output']
-            return result_ref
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for the PR. That is definitely a use case `vcstool` should cover.

The current patch breaks the export when you e.g. checked out a tag. I will look into it soon.

jbohren:
> The current patch breaks the export when you e.g. checked out a tag. I will look into it soon.

Ah yeah, that's a good point. I just added a commit that should determine:

* remote branch name
* remote tag name (find some remote with a tag referencing this sha)
* remote sha (find some remote with this sha)

If it can't determine the remote ref, it returns an error.

I'll squash this all down when you're ready.

dirk-thomas:
The added code does a lot of output parsing which tends to be fragile. I will wrap my head around it soon if there is an "easier" way to achieve the same.

jbohren:
> The added code does a lot of output parsing which tends to be fragile. I will wrap my head around it soon if there is an "easier" way to achieve the same.

Yeah the only way around that is probably [pygit2](https://github.com/libgit2/pygit2)

NikolausDemmel:
Ah, I didn't see this PR. Great stuff!

The parsing logic still needs to handle some more cases:

```
  quux:
    type: git
    url: ...
    version: master: ahead 50
  quax:
    type: git
    url: ...
    version: foo: ahead 243, behind 8] [Foo bar baz

```

jbohren:
That's an interesting case. I feel like in the case that you have commits which aren't on the remote branch, VCS export should choose the right branch, but give a warning.

@dirk-thomas What's the preferred way for issuing a warning to the user about this?

jbohren:
@NikolausDemmel Give 65103c2 a shot.

@dirk-thomas I documented the regex in the latest commit if you're still willing to accept output-parsing:
https://github.com/jbohren-forks/vcstool/blob/65103c24dcdf78f5956ed030a6c91e1e8c48c670/vcstool/clients/git.py#L87

dirk-thomas:
I just looked a little bit close into this ticket. I am very hesitant to use any kind of complex output parsing since it is *very* fragile. I thought about alternative steps to gain the necessary information and would propose doing the following instead:

* Determine if the wc has a specific branch checked out or is detached

  ```
  git rev-parse --abbrev-ref HEAD
  # output is either a branch name or "HEAD" when detached
  ```

* Depending on the exact flag there are two paths, first if `not exact`:

  * Determine the remote of the current branch:

    ```
    git rev-parse --abbrev-ref @{upstream}
    # output is the remote name
    ```

  * Get the url of the remote:

    ```
    git config --get remote.<REMOTE>.url
    ```

  * The result is the remote url + the branch name

* If `exact` is requested:

  * Determine the hash:

    ```
    git rev-parse HEAD
    ```

  * Get all remote names

    ```
    git remote
    ```

  * for each remote name check if the hash is part of the remote:

    ```
    git rev-list --remotes=<REMOTE>
    # search for <HASH> in the output
    ```

    * For the first successful remote the result is the remote url (as determine above) + the hash

This approach would not require any complex parsing. Does that sound reasonable?

dirk-thomas:
I have two more comments:

* while I do see the advantage of having a readable tag name in the export (instead of a hash when possible) I didn't implement this part earlier because I have seen people moving around tags (yeah, bad habbit!) and considered the hash to be more stable. May be we could make that an option?

* for the algorithm in this PR as well as for my proposed alternative it would be better if the selection of the remote is not only based on the alphabetical order. Instead we could check "known" / common remote names before others. I am thinking especially about `origin` and `upstream`.

jbohren:
> This approach would not require any complex parsing. Does that sound reasonable?

Yeah that looks reasonable. 

Unfortunately, as it stands, this version does what I need. I can't say I'll have any time to re-implement it  the way you described.

dirk-thomas:
Sure, I completely understand.

After figuring out all the details for the abstract algorithm described above the actual implementation shouldn't take much time. I will try to come up with another PR in the next days.

dirk-thomas:
I implemented the proposed logic and it seems to work for me: https://github.com/dirk-thomas/vcstool/tree/export_multiple_remotes

It would be great if you could give it a try and check if it works in your case too.

jbohren:
Yeah this looks like it works. It would be nice to still have a warning if there are any commits in the local branch which aren't on the remote that's being exported, but don't feel like you have to waste any time on that.

dirk-thomas:
Closing this in favor of #11.

I did not implement the warning since I am not confident it can be done reliable in all possible scenarios. If it warns only in some cases users might rely on the check and then the tools fails the expectations.

Also when using `--exact` it ensures that the specific current hash is actually available. When tracking a branch only various situations could happen where another user could not fetch the same state which was available on the machine `export` was invoked on.

jbohren:
@dirk-thomas Yeah, that's reasonable re:warning. Thanks for following up on this feature.

PR #9, opened 2015-02-11
custom: adding convenient entry points for custom --args subcommand

jbohren:


setup.py
```diff
@@ -51,6 +51,10 @@
             'vcs-push = vcstool.commands.push:main',
             'vcs-remotes = vcstool.commands.remotes:main',
             'vcs-status = vcstool.commands.status:main',
+            'vcs-git = vcstool.commands.custom:git_main',
+            'vcs-hg = vcstool.commands.custom:hg_main',
+            'vcs-svn = vcstool.commands.custom:svn_main',
+            'vcs-bzr = vcstool.commands.custom:bzr_main',
```
dirk-thomas: Please add those line in alphabetical order.

vcstool/commands/custom.py
```diff
@@ -72,6 +72,17 @@ def main(args=None):
     output_results(results, hide_empty=args.hide_empty)
     return 0
 
+def git_main(args=None):
+    return main(['--git','--args']+(args if args else sys.argv[1:]))
```
dirk-thomas: The condition is to unspecific. If the function gets invoked e.g. with an empty list it falls back to sys.argv. Please compare with `is not None`.

For readability can you also please move setting the args before the return:

    if args is None:
        args = sys.argv[1:]

Please add a space after the comma and around the plus operator (PEP8) - same below.

dirk-thomas:
Please also add the corresponding scripts which makes it easier to use this package from source.

jbohren:
Feedback issues resolved.

scripts/vcs-bzr
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env python
+
+import sys
+from vcstool.commands.custom import bzr_main
+
+sys.exit(git_main() or 0)
```
dirk-thomas: Calling the wrong function - same below.

vcstool/commands/custom.py
```diff
@@ -72,6 +72,25 @@ def main(args=None):
     output_results(results, hide_empty=args.hide_empty)
     return 0
 
+def bzr_main(args=None):
+    if args is None:
+        args = sys.argv[1:]
+    return main(['--bzr', '--args']+args)
```
dirk-thomas: Please add spaces around the plus operator - same below.

jbohren:
Dumb typos resolved.

dirk-thomas:
Great - thank you!

PR #11, opened 2015-03-02
fix export with multiple remotes

dirk-thomas:
Replaces #8.

It fixes the export command when the repo has multiple remotes without requiring complex parsing of the git output.

Additionally when using `--exact` the patch will prefer `origin` / `upstream` remotes when the hash is available for them. That is a "better" behavior than picking the first available remote (in alphabetical order) containing the hash.

@jbohren @NikolausDemmel FYI

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -101,7 +101,13 @@ class GitClient(VcsClientBase):
                 return result_remotes
             remotes = result_remotes['output'].splitlines()
 
-            # TODO prefer original / upstream remote
+            # prefer origin and upstream remotes
+            if 'upstream' in remotes:
+                remotes.remove('upstream')
+                remotes.insert(0, 'upstream')
+            if 'origin' in remotes:
+                remotes.remove('origin')
+                remotes.insert(0, 'origin')
 
             # for each remote name check if the hash is part of the remote
             for remote in remotes:
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -33,61 +33,139 @@ class GitClient(VcsClientBase):
         return self._run_command(cmd)
 
     def export(self, command):
-        result_url = self._get_url()
-        if result_url['returncode']:
-            return result_url
-        url = result_url['output'][0]
+        exact = command.exact
+        if not exact:
+            # determine if a specific branch is checked out or ec is detached
+            cmd_branch = [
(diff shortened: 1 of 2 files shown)


jbohren:
+1

NikolausDemmel:
Thanks Dirk! I can test only next week, though.

dirk-thomas:
@NikolausDemmel I will go ahead with the merge (since it seems to work for two of us and is otherwise broken annyway). I will also release a new version (to clearly mark the Python 2 and 3 packages to conflict with each other). If the new version doesn't work for you when you have a chance to try it please comment here.

NikolausDemmel:
@dirk-thomas: The latest version fixes the issue for some cases, but for one repo where I'm currently on a branch that does not have its upstream set, it fails with:

    Could not determine ref: fatal: No upstream configured for branch 'bluebotics-ros'

Note that this git repo has 2 remotes, only one of which has a branch matching the current local branch by name.

I guess the desired behaviour would be to pick a remote that matches, but maybe still warn the user about this use of a "heuristic" on stderr.

dirk-thomas:
I am not sure if I understand your use case correctly. Is the following interpretation correct?

* you have created a local branch with, e.g. `git branch my_branch`
* you have pushed it to some remote, e.g. `git push origin my_branch`
* and now you try to export?
* then it is further true that also an invocation of `git pull` would fail telling you to use `git branch --set-upstream my_branch origin/my_branch` to set the remote?

NikolausDemmel:
This is correct.

I guess this is a special case and maybe it's best to simply not handle it. The error message could be improved though, informing the user about which repository it was that failed.

dirk-thomas:
Thanks for clarifying. Since git does not know which remote the actual upstream is I would rather not let vcstool make any guesses. The user can easily solve the uncertainty by either setting the upstream explicitly or start using `git push -u origin my_branch` when pushing new branches to also set the upstream.

I expected the error message to be in the repository specific block but since the `export` command directly outputs yaml that is obviously not the case.

I have created another PR (#13) which makes it still output the correct repositories. But below it will print all errors (and warnings) below the yaml output. Can you please give it a try and see if it is sufficient for your use case?

NikolausDemmel:
This makes sense. I agree that improved error message as in #13 should be sufficient for now. Thanks!

PR #12, opened 2015-03-11
Windows support

wjwwood:
This:

- Disables color output, since escape sequences do not work on Windows.
- Change `find_executable` to consider files with `.exe`, e.g. `git.exe` for `git`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 *.pyc
+build
 deb_dist
 dist
 vcstool.egg-info
-build
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 deb_dist
 dist
 vcstool.egg-info
+build
--- a/vcstool/clients/vcs_base.py
+++ b/vcstool/clients/vcs_base.py
@@ -41,11 +41,26 @@ class VcsClientBase(object):
         return None
 
 
-def find_executable(file_name):
+def is_executable_file(file_path):
+    return os.path.isfile(file_path) and os.access(file_path, os.X_OK)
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 deb_dist
 dist
 vcstool.egg-info
+build
```
dirk-thomas: Nitpick: can you please move this up to be in alphabetical order?

vcstool/clients/vcs_base.py
```diff
@@ -41,11 +41,26 @@ def _create_path(self):
...
+            check_for_exe = False
     for path in os.getenv('PATH').split(os.path.pathsep):
         file_path = os.path.join(path, file_name)
-        if os.path.isfile(file_path) and os.access(file_path, os.X_OK):
+        if is_executable_file(file_path):
             return file_path
+        if check_for_exe:
+            file_path += '.exe'
+            if is_executable_file(file_path):
+                return file_path
```
dirk-thomas: What about changing it to the following snippet to keep the logic in one place:

    file_paths = [os.path.join(path, file_name)]
    if check_for_exe:
        file_paths.append(file_paths[0] + '.exe')
    for file_path in file_paths:
        if os.path.isfile(file_path) and os.access(file_path, os.X_OK):
            return file_path

dirk-thomas:
I cherry-picked some of the changes, implemented some a bit dfferent and integrated `shutil.which` like `osrf_pycommon` to get rid of the custom `find_executable` implementation. Thanks!

PR #13, opened 2015-03-13
improve output of export command in case of errors

dirk-thomas:
Further improve #11.

### DIFF
--- a/vcstool/commands/export.py
+++ b/vcstool/commands/export.py
@@ -28,27 +28,46 @@ def get_parser():
 
 
 def output_export_data(result, hide_empty=False):
+    # errors are handled by a separate function
+    if result['returncode']:
+        return
+
+    path = get_path_of_result(result)
+
+    try:
+        lines = []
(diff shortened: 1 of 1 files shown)


NikolausDemmel:
Cool. Thanks for the quick response.

Example output:

```
$ vcs export
repositories:
  Framework:
    type: git
    error: "Could not determine ref: fatal: No upstream configured for branch 'foo'"

Framework: Could not determine ref: fatal: No upstream configured for branch 'foo'
```

Did you intend for the error to be printed twice?

In any case. For me this is more than good enough. Cheers!

dirk-thomas:
I was still shuffling around code on this PR - now it should be final and only print errors once.

NikolausDemmel:
+1

Looks good:
```
$ vcs export
repositories:
Framework: Could not determine ref: fatal: No upstream configured for branch 'foo'
```
Thanks for implementing!

dirk-thomas:
Thanks for testing!

PR #17, opened 2015-08-13
add short option for --hide-empty

codebot:
-h was taken by the obvious use, so this suggestion uses -e instead

I guess we could do capital-h  ( -H ) but i'm not sure what is more obvious.


dirk-thomas:
Why `-e` - is there any other program using it?

You can already use any shorter version which is unique, e.g. `--hi`. Do we really need a shorter option then that?

dirk-thomas:
If the option would be called `--skip-empty` we could use `-s`. That would also be similar to other tools using `-s` for summaries (basically only show the relevant stuff).

What do you think about adding `--skip-empty` as a synonym and `-s` as the short version?

dirk-thomas:
Thanks, squashed and merged in 176ec2381a403519f4b68af26a5c4293a082c448.

PR #29, opened 2016-08-05
Handle exporting repos with a a "rev-parse" prefixed with "remotes/"

dotjairo:
If the command `git rev-parse --abbrev-ref @{upstream}` in a repo returns `remotes/<remote>/<branch>`, then `vcs export` will print the error: `src/mycatkinpkg: Could not determine remote url:`

This pull request adds logic to handle the `remotes/` prefix in the command above. If the prefix is not present the logic is unchanged.

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -71,10 +71,16 @@ class GitClient(VcsClientBase):
 
             # determine remote
             suffix = '/' + branch_name
+            prefix = 'remotes/'
             assert branch_with_remote.endswith(branch_name), \
                 "'%s' does not end with '%s'" % \
                 (branch_with_remote, branch_name)
-            remote = branch_with_remote[:-len(suffix)]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you please a simple sequence of command to reproduce the use case.

dotjairo:
Hey Dirk, I'm actually unable to reproduce the issue in a new git repo. I'm OK closing this if we can't reproduce it

dirk-thomas:
After some more consideration I was able to identify your use case. You likely have a remote `foo` with a branch `bar` but also a local branch named `foo/bar`. `--abbrev-rev @{upstream}` would be ambiguous if it would return just `foo/bar`. Therefore it prepends the additional `remotes/` to make the result unambiguous.

I have committed a slightly different patch. The check for the prefix is separated form the check for the suffix: 20007bcecc6c5b16193cacbd4548a658c47a94d7

PR #34, opened 2017-03-03
don't show 'same repo' when using --hide

dirk-thomas:
Skip repos which show "same repo" message when passing `--hide`.


dirk-thomas:
Released in 0.1.22.

PR #36, opened 2017-03-20
do not hide git status when ahead or behind

### LINKED ISSUE #35: [feature] Make vcs status --hide show when branch diverged from their remote
If it is of any interest I can try to take a shot at it. For the [git client](https://github.com/dirk-thomas/vcstool/blob/master/vcstool/clients/git.py#L327-L340) for example, we could apply something like:
```python
-            cmd = [GitClient._executable, 'status', '-s']
+            cmd = [GitClient._executable, 'status', '-s', '-b']
             if command.quiet:
                 cmd += ['--untracked-files=no']
             result = self._run_command(cmd)
-            if result['returncode'] or not result['output']:
+            if result['returncode'] or ('\n' not in result['output'] and 'ahead' not in result['output'] and 'behind' not in result['output']):
+                result['output'] = ''
```
Not sure how it would work for locales though. Didn't look at other vcs systems to see if something similar can be done easily

dirk-thomas:
Fixes #35

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -326,13 +326,26 @@ class GitClient(VcsClientBase):
 
     def status(self, command):
         self._check_executable()
-        if command.hide_empty:
+        while command.hide_empty:
+            # check if ahead
+            cmd = [GitClient._executable, 'log', '@{push}..']
+            result = self._run_command(cmd)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Released in 0.1.23.

PR #37, opened 2017-04-27
add --force option to import verb

dirk-thomas:
If a different repository exists in a directory the `--force` option can be used to remove it before cloning the new repository. Without the option the command failed for that repository.


dirk-thomas:
Released in 0.1.24.

PR #51, opened 2017-08-09
support zip

### LINKED ISSUE #48: ZIP file support
Please add a handler for zip files (similar to the tar file handler). Thanks.

dirk-thomas:
Fixes #48.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -80,7 +80,9 @@ Usually the data of a previously exported file is piped in::
   vcs import < my.repos
 
 The ``import`` command also supports input in the `rosinstall file format <http://www.ros.org/doc/independent/api/rosinstall/html/rosinstall_file_format.html>`_.
-Only for this command vcstool supports a pseudo client ``tar`` which fetches a tarball from an URL and unpacks its content.
+
+Only for this command vcstool supports the pseudo clients ``tar`` and ``zip`` which fetch a tarball / zipfile from an URL and unpack its content.
+The version key should in this case be the folder name the archive unpacks all its content into.
 
 
 Advanced features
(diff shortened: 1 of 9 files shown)


BrannonKing:
I executed the tool in the zip branch. It says that it is downloading and extracting the zip file successfully, but when I look in the destination folder after completion there are no files there.

dirk-thomas:
The version value is probably incorrect. Can you share the URL and version you used?

BrannonKing:
vcs --version outputs `vcs 0.1.28`. The zip I downloaded is private to the company. I just tried it on this one:
```
repositories:
  test_dest:
    type: zip
    url: http://www.7-zip.org/a/7za920.zip
    version: 920
```
 and got this error:
```
Invocation of command 'import' on client 'zip' failed: TypeError: unsupported operand type(s) for +: 'int' and 'str' (/usr/local/lib/python3.5/dist-packages/vcstool-0.1.28-py3.5.egg/vcstool/clients/zip.py:91)
```

dirk-thomas:
Please try it with `version: 7za920` since that is the folder the zip file would normally unpack its content into.

I addressed the `TypeError` in 0d74ab0bd95a9a48df6601b01898240e40cdfd21 in case someone needs a non-string version value.

BrannonKing:
I'm seeing nothing in the destination folder. Sorry. I'm running on Ubuntu 16.04 + updates.

dirk-thomas:
Ok, I see. The ones I tried the feature with were all consistent with the existing tar feature and used a specific folder for all the content inside the archive (e.g. https://github.com/dirk-thomas/vcstool/pull/51/files#diff-5a86edcb72f7cff1c871e1292dee2396R12). In your example that is not the case and the version key doesn't make any sense. I updated this PR to make the version key optional for `tar` and `zip`. In your case you should remove the version key and it should work as desired.

BrannonKing:
Ship it. Thanks.

dirk-thomas:
Thanks for the feedback.

dirk-thomas:
This will soon be available as 0.1.29.

PR #53, opened 2017-09-10
add --retry option to import verb

### LINKED ISSUE #52: Add option to `import` to retry clone of repositories until success or timeout
Some ci builds fail while using vcs to download dependencies because of temporary network issues ([example](http://ci.ros2.org/view/nightly/job/nightly_linux_debug/577/console)). 
 
```
07:36:31 === src/eProsima/Fast-RTPS (git) ===
07:36:31 Could not clone repository 'https://github.com/eProsima/Fast-RTPS.git': Cloning into '.'...
07:36:31 error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.
07:36:31 fatal: The remote end hung up unexpectedly
07:36:31 fatal: early EOF
07:36:31 fatal: index-pack failed
```

I suspect this would had succeeded if vcs waited a few seconds and tried again. Adding an option to `import` which would retry cloning with some delay or increasing backoff may reduce the number of failing jobs on the build farm.

dirk-thomas:
Fixes #52.

### DIFF
--- a/test/status.txt
+++ b/test/status.txt
@@ -8,4 +8,5 @@ nothing to commit, working tree clean
 === ./vcstool (git) ===
 On branch master
 Your branch is up-to-date with 'origin/master'.
+
 nothing to commit, working tree clean
--- a/test/test_commands.py
+++ b/test/test_commands.py
@@ -85,9 +85,13 @@ class TestCommands(unittest.TestCase):
 
     def test_status(self):
         output = run_command('status')
-        # replace message from git < 2.9.1
+        # replace message from older git versions
+        # https://github.com/git/git/blob/3ec7d702a89c647ddf42a59bc3539361367de9d5/Documentation/RelNotes/2.10.0.txt#L373-L374
         output = output.replace(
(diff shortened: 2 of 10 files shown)


sloretz:
@dirk-thomas This appears to work. I tested by disabling/enabling networking multiple times during an `import` with `--retry 5` using a repos file with ~10 repos. I never disabled networking for more than 5 seconds, but all the repo's were cloned successfully.

An enhancement to this enhancement would be to print something when a retry is occurring, but that only aids testing. I doubt most users would care if there was a retry as long as the import was successful.

dirk-thomas:
@sloretz Thank you for testing it.

I updated the default retry to be `2` in a1796208bbff956d3836044e80ac62af3339a04f. That value is now being used for both loading data from urls as well as the vcs commands using network access.

Since the progress indicator is currently fully asynchronous it would be more difficult to add a message for this. I will leave it out for now hoping that users don't need to be bothered "with the details" :wink:

dirk-thomas:
:ship: in 0.1.30.

PR #61, opened 2018-01-23
use pytest on Travis

dirk-thomas:
Use `pytest` to run the tests since it provides much better introspection in case tests fail.


dirk-thomas:
The second commit addresses a change in `git` which as of version 2.15.0 changed the output of some commands.

PR #66, opened 2018-03-28
Avoid unnecessary failure when a git repo has multiple remotes

stonier:
As noted in the comment, this fixes a hard failure right now when you try to import on top of an existing repo with multiple remotes, but could be made more robust by checking off against each remote in the list instead of just the first.

Example of how it breaks right now:

```
$ vcs import src < src/delphyne_gui/delphyne.repos
=== src/delphyne (git) ===
Could not determine remote url: error: invalid key (newline): remote.origin
stonier.url
=== src/delphyne_gui (git) ===
Could not determine remote url: error: invalid key (newline): remote.origin
stonier.url
```

In those repos:

```
./src/delphyne$ git remote show
origin
stonier
```

With some extra debugging to show what is getting captured by `_get_url`:

```
$ vcs import src < src/delphyne_gui/delphyne.repos
Running command ['/usr/bin/git', 'config', '--get', u'remote.origin\nstonier.url']
Running command ['/usr/bin/git', 'config', '--get', u'remote.origin\nstonier.url']
=== src/delphyne (git) ===
Could not determine remote url: error: invalid key (newline): remote.origin
stonier.url
=== src/delphyne_gui (git) ===
Could not determine remote url: error: invalid key (newline): remote.origin
stonier.url
```

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -187,12 +187,13 @@ class GitClient(VcsClientBase):
         self._check_executable()
         if GitClient.is_repository(self.path):
             # verify that existing repository is the same
-            result_url = self._get_url()
-            if result_url['returncode']:
-                return result_url
-            url = result_url['output'][0]
-            remote = result_url['output'][1]
-            if url != command.url:
+            result_urls = self._get_remote_urls()
+            if result_urls['returncode']:
+                return result_urls
+            for url, remote in result_urls['output']:
+                if url == command.url:
+                    break
+            else:
                 if not command.force:
                     return {
                         'cmd': '',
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Using just the first remote is not going to work in many cases. I checked a few local working copies and it appears that the remotes are shown in alphabetical order. An in a lot of cases `origin` is not the first so I think this patch needs to consider all remotes.

vcstool/clients/git.py
```diff
@@ -281,24 +284,26 @@ def import_(self, command):
...
+            url = 'n/a' if result_url['returncode'] else result_url['output']
+            remote_urls.append(
+                {
+                    'cmd': ' && '.join([result_remote['cmd'],
+                                        result_url['cmd']]),
+                    'cwd': self.path,
+                    'output': [url, remote],
+                    'returncode': result_url['returncode']
+                })
+        return remote_urls
```
dirk-thomas: Maybe it would be easier to return only a result rather than a list of results? The `output` of the single result could still be a list of url-remote-tuples. This would make the error handling in the calling location much easier.

vcstool/clients/git.py
```diff
@@ -281,23 +282,26 @@ def import_(self, command):
...
+            if not result_url['returncode']:
+                remote_urls.append((result_url['output'], remote))
         return {
-            'cmd': ' && '.join([result_remote['cmd'], result_url['cmd']]),
+            'cmd': cmd,
             'cwd': self.path,
-            'output': [url, remote],
-            'returncode': 0
+            'output': remote_urls if remote_urls else
+                'Could not determine any of the remote urls',
```
dirk-thomas: The linter test currently fails for this line.

dirk-thomas:
Thank you for this improvement.

Issue #69, opened 2018-05-09
VCS tool 0.1.35 failing for svn repos


asibryant:
I'm trying to checkout svn repos with vcs tool 0.1.35 and they're failing with the error:
```
Invocation of command 'import' on client 'svn' failed: TypeError: %d format: a number is required, not str (C:\Python36\lib\site-packages\vcstool\clients\svn.py:128)
```
The same repos file works fine with vcstool 0.1.29

dirk-thomas:
I don't see any related changes between these two releases. Can you please share a reproducible example to figure out why it has changed for you.

asibryant:
It doesn't seem to have problems on linux, but on windows with version 0.1.35 it will have the problem on this repos file.

```
repositories:
  stuff:
    type: svn
    url: https://svn.eionet.europa.eu/repositories/airquality/
```

dirk-thomas:
Hm , that looks weird. If the entry doesn't have a version key I would expect the `version` field in the `ImportCommand` to be `None`: https://github.com/dirk-thomas/vcstool/blob/b6d5f5f2e7be6519d23420e8f677dbdae0f58839/vcstool/commands/import_.py#L141-L144). And if the version is `None` the line of code the `TypeError` is happening shouldn't even be used due to the condition it is being wrapped in (see https://github.com/dirk-thomas/vcstool/blob/b6d5f5f2e7be6519d23420e8f677dbdae0f58839/vcstool/clients/svn.py#L127-L128).

Maybe you can add some debug statement before that line and report what type and value `command.version` is on your case.

You can also try the following patch (https://github.com/dirk-thomas/vcstool/commit/e3c95ab2218f18c0f61bcb08502db4578cf91e17) and see if it helps.

asibryant:
It seems that my example was invalid. Here is a better one.
```
repositories:
  stuff:
    type: svn
    url: https://svn.alfresco.com/repos/alfresco-open-mirror/alfresco
    version: HEAD
```
Sorry for that. The trick here is that the version is always parsed as a string, even if it is an integer, so the print statement trying to print is a number will always fail.

dirk-thomas:
Thanks for clarifying. That example makes sense.

asibryant:
What are the chances of getting a release for these changes? It would just be nice to install them from pypi.

dirk-thomas:
Absolutely. :ship: 0.1.36

PR #70, opened 2018-05-11
make hg use non interactive

asibryant:
hg wasn't using noninteractive, so a computer without keyring set up would just hang.


asibryant:
#69

dirk-thomas:
Please don't mix unrelated changes in a single PR. The change related to the SVN version is already in  branch which I linked to in https://github.com/dirk-thomas/vcstool/issues/69#issuecomment-388202843. I just created a PR from that branch: #71. Please remove the SVN version related commit from this PR.

dirk-thomas:
Thank you for the patch.

PR #80, opened 2019-01-30
Add command line argument --no-output to suppress output_results() call

herrfrei:
I use vcstool in an automatic build environment and want to suppress the VCS tool output.

### DIFF
--- a/vcstool/commands/command.py
+++ b/vcstool/commands/command.py
@@ -45,7 +45,7 @@ def add_common_arguments(
         '--debug', action='store_true', default=False,
         help='Show debug messages')
     group.add_argument(
-        '-q', '--quiet', action='store_true',
+        '--no-output', action='store_true',
         default=False, help='Be quiet, no output from VCS tool execution')
     if not skip_hide_empty:
         group.add_argument(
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I am wondering why you need an option for this behavior. How about redirecting the output if you don't want to see it: `> /dev/null`?

dirk-thomas:
May I ask what automatic build environment you are using? E.g. in the case of something like Travis CI I would suggest to instead fold the output of `vcstool` to not be in the way but available in case the output needs to be investigated.

Do you want to also suppress the progress output (which the patch currently doesn't)? How about potential error messages?

> I use vcstool as a library

I am wondering if an API would be nicer for this use case... Maybe adding optional arguments to the `main()` of all verbs to pass a custom `stdout` / `stderr`? That would:
* avoid a command line option which shouldn't have a need to be used on the command line
* support your use case of dropping all output
* handle `stdout` and `stderr` separately (e.g. only drop the former)

herrfrei:
It's our company-internal build system (self-made, Python and CMake based, platform independent). Error messages shouldn't be suppressed of course, so my fix seems to be only a dirty workaround...
Your idea of adding custom streams for stdout / stderr to the main() functions seems to be a elegant way to solve my problem!

dirk-thomas:
Please see #83 which introduces the two keyword arguments `stdout` and `stderr` to the `main` function (for now only for the `pull` verb). Please let me know if this works for your use case. If yes, it can be extended to all verbs.

herrfrei:
Dirk, many thanks for this (much more extensive) modification, it works very well. I close this request now.

PR #90, opened 2019-03-29
Adding validate command.

JWhitleyAStuff:
The `validate` command simply re-uses some of the functions from `import_.py` to validate the contents of a `.repos` file in vcstool or rosinstall format. This is useful for validating hand-modified files or for CI validation of vcstool-based workspace repos.


JWhitleyAStuff:
Adding more functionality.

JWhitleyAStuff:
The `validate` command now works for all client types except bzr (there is no "check repo existence/validity" command for Bazaar that I can find).

dirk-thomas:
Before diving into the actual patch can you please clarify the scope of the new `validate` command and what is supposed to check for. A few things which come to my mind: well formed input file, referenced URLs exist / are fetchable, referenced versions exists.

JWhitleyWork:
Hey, Dirk. So, you have the general idea correct regarding the functionality. There is just one exception: I have not currently found a way to test to see if the referenced version exists on the remote repo. So, here is the current feature list for the `validate` keyword that I've been working off of in my head (mostly to support https://github.com/autowarefoundation/autoware_ai/issues/640):

1. Validate that the .repos or .rosinstall file is valid YAML syntax.
2. Validate that each endpoint URI exists and is reachable.
3. Validate that the endpoint is the correct type of repo or file as provided in the .repos or .rosinstall file.
4. Do all of the above without pulling the contents of the remote repo or file.

dirk-thomas:
> I have not currently found a way to test to see if the referenced version exists on the remote repo.

For specific hashes which are not HEAD of any of the branches I am not aware of a way to do that (purely focusing in git). But `ls-remote` will give you all branch and tag names which you could use to validate. And maybe print a message / warning if a hash is given which is not a HEAD.

JWhitleyAStuff:
@dirk-thomas - Since this only applies to git and is not always reliable (as you mentioned, only works if the hash is HEAD on a branch which, I would assume, is usually not the case since the intent to tying to a hash is usually to avoid new feature additions/changes), are you OK with accepting the PR in it's current form or do you want me to attempt adding this feature first?

JWhitleyAStuff:
It actually looks like Mercurial handles this even better than `git`. Mercurial has the command `hg identify -r <REV> <URL>` which will identify if a revision (branch, tag, or hash) exists on the remote. I'll work on adding at least these two.

JWhitleyAStuff:
@dirk-thomas - I have added version-based validation to `hg` and `svn`. I added _pseudo-validation_ to `git` which will show a warning if the URL is valid but the contents of the `version:` attribute could not be verified. I also added a more comprehensive `list2.repos` file for testing the validate command. I'm not sure if you want to use this for the other tests as well but it contains one entry of each type and a `version:` tag of each type (hash, tag, and branch) for those that support them.

JWhitleyAStuff:
@dirk-thomas (I know you're super busy - just wanted to hit you up with a *ping* in case you forgot)

README.rst
```diff
@@ -93,6 +93,18 @@ Only for this command vcstool supports the pseudo clients ``tar`` and ``zip`` wh
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``vcs validate`` command takes a YAML file which is passed in via stdin and validates its contents and format.
+The data of a previously-exported file or hand-generated file are piped in::
+
+  vcs validate < my.repos
+
+The ``validate`` command also supports input in the `rosinstall file format <http://www.ros.org/doc/independent/api/rosinstall/html/rosinstall_file_format.html>`_.
+
+If a file fails to validate, an error will be printed and the command will exit. If the file passes validation, the response "Validation succeeded!" will be printed.
```
dirk-thomas: See other comment about the difference of the `validate` output to the other existing commands.

It also signals the overall result with its return code.

test/test_commands.py
```diff
@@ -134,6 +135,12 @@ def test_reimport(self):
         # newer git versions don't append three dots after the commit hash
         assert output == expected or output == expected.replace(b'... ', b' ')
 
+    def test_validate(self):
+        output = run_command(
+            'validate', ['--input', REPOS2_FILE, '.'])
```
dirk-thomas: The new `validate` command doesn't need / use the positional `path` argument and therefore shouldn't have it.

test/validate.txt
```diff
@@ -0,0 +1,14 @@
...
+VALID immutable/git_hash (git) 
+Remote repository validated but requested version is not ref so unable to verify.
+VALID immutable/git_tag (git) 
+VALID immutable/hg_branch (hg) 
+VALID immutable/hg_hash (hg) 
+VALID immutable/hg_tag (hg) 
+VALID immutable/svn_ver (svn) 
+VALID immutable/tar_hash (tar) 
+VALID immutable/zip_hash (zip) 
+Endpoint validation succeeded!
```
dirk-thomas: The way `validate` outputs information doesn't match the structure / style used by the other commands.

vcstool/clients/git.py
```diff
@@ -395,6 +395,86 @@ def status(self, command):
...
+                    elif hash_and_ref[1].find('refs/heads/') > -1:
+                        refs.append(hash_and_ref[1][11:])
+                    else:
+                        refs.append(hash_and_ref[1])
+
+            # test for refs first
+            for _ref in refs:
+                if command.version == _ref:
+                    ref_found = True
+                    break
```
dirk-thomas: The logic to determine `ref_found` seems overly complicated.

vcstool/clients/git.py
```diff
@@ -395,6 +395,86 @@ def status(self, command):
...
+            refs = []
+            ref_found = False
+            output_lines = result_ls_remote['output'].splitlines()
+
+            for line in output_lines:
+                hash_and_ref = line.split()
+                hashes.append(hash_and_ref[0])
+
+                # ignore pull request refs
+                if hash_and_ref[1].find('refs/pull/') == -1:
```
dirk-thomas: The ref has to actually start with `refs/...`, no? In that case instead of using `find` `startswith` makes more sense.

Same below.

vcstool/clients/git.py
```diff
@@ -395,6 +395,86 @@ def status(self, command):
...
+
+        self._check_executable()
+
+        cmd_ls_remote = [GitClient._executable, 'ls-remote']
+        cmd_ls_remote += ['-q', '--exit-code']
+        cmd_ls_remote += [command.url]
+        result_ls_remote = self._run_command(
+            cmd_ls_remote,
+            retry=command.retry,
+            env={'GIT_TERMINAL_PROMPT': '0'})
```
dirk-thomas: The existing `os.environ` needs to be extended to not wipe out other essential environment variables.

vcstool/clients/git.py
```diff
@@ -395,6 +395,86 @@ def status(self, command):
...
+            if not ref_found:
+                for _hash in hashes:
+                    if _hash.find(command.version) == 0:
+                        ref_found = True
+                        break
+
+            if not ref_found:
+                cmd = result_ls_remote['cmd']
+                output = 'Remote repository validated but requested '
+                output += 'version is not ref so unable to verify.'
```
dirk-thomas: The information for each message should include more context like the URL.

Same below.

vcstool/clients/git.py
```diff
@@ -395,6 +395,86 @@ def status(self, command):
...
+                output = result_ls_remote['output']
+        else:
+            cmd = result_ls_remote['cmd']
+            output = result_ls_remote['output']
+
+        return {
+            'cmd': cmd,
+            'cwd': self.path,
+            'output': output,
+            'returncode': 0
```
dirk-thomas: Returning `None` instead of `0` allows entries which have been successfully validated to be skipped with `--hide-empty`.

vcstool/commands/validate.py
```diff
@@ -0,0 +1,154 @@
...
+                    file=stdout)
+
+
+def output_results(results, output_handler=output_result, hide_empty=False):
+    # output results in alphabetic order
+    path_to_idx = {
+        result['command'].real_path: i for i, result in enumerate(results)}
+    idxs_in_order = [path_to_idx[path] for path in sorted(path_to_idx.keys())]
+    for i in idxs_in_order:
+        output_handler(results[i], hide_empty=hide_empty)
```
dirk-thomas: Instead of generating completely custom output reuse the existing output result functions.

vcstool/commands/validate.py
```diff
@@ -0,0 +1,154 @@
...
+        return 1
+
+    print('Format validation succeeded!')
+    print('Validating endpoints...')
+
+    jobs = generate_jobs(repos, args)
+
+    results = execute_jobs(
+        jobs, number_of_workers=args.workers,
+        debug_jobs=args.debug)
```
dirk-thomas: For larger sets of repos supporting the `show_progress` option is useful.

JWhitleyAStuff:
Thanks for getting to this. Everything in your feedback looks spot-on. Some of what you replaced has a lot to do with me not being as fluent in Python as I am in C++ so I appreciate the help. I'm good with the state of the PR if you are!

dirk-thomas:
I added one more minor change to consolidate the test repos. The second repos file only contains additional hg/svn repos and the test runs for both: a0801a4cad266a1124ec9fb1e8d77388d9056a44.

vcstool/clients/svn.py
```diff
@@ -209,6 +209,56 @@ def status(self, command):
...
+                SvnClient._executable, 'info',
+                command.url + "@" + command.version]
+            result_info_ver = self._run_command(
+                cmd_info_ver,
+                retry=command.retry)
+
+            if result_info_ver['returncode']:
+                result_info_ver['output'] = \
+                    "Specified version not found on remote repository" + \
+                    "'%s':'%s' : %s" % (command.url, result_info_ver['output'])
```
dirk-thomas, line 244: The string has 3 placeholders but only 2 substitutions. Fixed in #133.

PR #95, opened 2019-06-10
Set path_nargs to ? for simple_main.

clalancette:
This ensures that the commands the rely on simple_main (like
log) continue to get the old behavior of a single list of
paths (not a list of list of paths).

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

### DIFF
--- a/vcstool/commands/command.py
+++ b/vcstool/commands/command.py
@@ -83,7 +83,7 @@ def existing_dir(path):
 
 
 def simple_main(parser, command_class, args=None):
-    add_common_arguments(parser)
+    add_common_arguments(parser, path_nargs='?')
     args = parser.parse_args(args)
 
     command = command_class(args)


clalancette:
I'm still testing this, but at the very least this fixes the `log` command as pointed out in https://github.com/ros2/build_cop/issues/215

clalancette:
All right, this seems to pass my local tests.  The `test-validate` test always fails (locally and in Travis) because the webserver at red-bean.com seems to be offline.

clalancette:
> Can you link to a passing build?

Unfortunately, it's not that easy to do this on http://ci.ros2.org.  We don't have inbuilt facilities for checking out a custom vcstool version.  While I could write something, this is clearly wrong as it is, so I'd prefer to just pass the unit/local tests here and do a new release.

vcstool/commands/command.py
```diff
@@ -83,7 +83,7 @@ def existing_dir(path):
 
 
 def simple_main(parser, command_class, args=None):
-    add_common_arguments(parser)
+    add_common_arguments(parser, path_nargs='?')
```
dirk-thomas: This is not the right way to address the regression. Before `simple_main` was using the default value `single_path=False` and the semantic of the default hasn't changed: `path_nargs='*'`.

dirk-thomas:
Closing in favor of #96.

PR #96, opened 2019-06-10
fix regressions with path/paths arguments

dirk-thomas:
Replaces #95.

Fixes regression introduced in https://github.com/dirk-thomas/vcstool/pull/90/files#diff-ab1fa1c2a07801d2a7ab588a990f0879L72

Instead of adding the argument `path` or `paths` with its string or list of strings default value the patch introducing the regression always added a `path` argument.

### DIFF
--- a/test/list2.repos
+++ b/test/list2.repos
@@ -13,5 +13,5 @@ repositories:
     version: icedtea-3.0.0
   svn/rev:
     type: svn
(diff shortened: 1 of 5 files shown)


hidmic:
> the patch always adds a path argument.

I see that it's either depending `path_nargs` value?

dirk-thomas:
I was referring to the patch which introduced the regression.

PR #99, opened 2019-07-08
add option --skip-existing to not change checked out branch when no version is specified

mxgrey:
When the `version:` field is left blank for `git` repositories, the `GitClient.import_` function will currently use `git remote show ...` to find the name of the `HEAD` branch of the remote repository and then automatically check out that branch.

This is undesirable if `vcstool` is being used to clone/update the source directories in a workspace where the local version might want a different branch than the remote's `HEAD`. To get that behavior, this PR introduces an `--ignore-head` flag that will tell the `GitClient.import_` function to not perform any checkout if no version is explicitly listed.

From glancing over the other clients, it looks like this new behavior will actually make the `GitClient` behave *more* similarly to the other clients. But since it's a potentially significant change in behavior, this PR introduces a new flag to explicitly enable this behavior, and `vcs-import` will continue to default to the old behavior.


dirk-thomas:
I am not sure the behavior you describe is desired.

In general the idea is that a `.repos` file specifies a set of repositories with a specific branch/tag/hash. After an initial import of the file the file system should contain exactly the described state. An empty version means the default branch in this context (similar to the git command line interface). If you update existing working copies which might be in any arbitrary state the goal is that after running `vcs import` your file system is in the same state as it would be after an initial / fresh import.

In your use case that doesn't seem to be the case / desire. While the `.repos` file implicitly references the default branch you want the working copy to stay on an already checked out custom branch instead. I would consider this undesired behavior of the tool and therefore I am not sure if adding such an option is a good idea.

I don't know anything about your use case but why are the branch names in the `.repos` file and the local working copy different in the first place? Maybe a change in process would be an alternative.

mxgrey:
> After an initial import of the file the file system should contain exactly the described state.

When the state is not described (i.e. no version is specified), then what should be considered the exact described state?

> An empty version means the default branch in this context

That's not currently the behavior for the [mercurial client](https://github.com/dirk-thomas/vcstool/blob/1f19788136b9af249140c3315458414f52c46bf5/vcstool/clients/hg.py#L161) which simply won't check out any particular branch if no `version` is specified. The git client seems to be unique in this behavior.

> If you update existing working copies which might be in any arbitrary state the goal is that after running vcs import your file system is in the same state as it would be after an initial / fresh import.

My hope is that we can allow developers (i.e. people writing the `.repos` files) to leave certain versions unspecified so that the working copy can be whatever the end users (i.e. people consuming the `.repos` files) currently have on their machine.

Conceptually I think this makes perfect sense, and I would have expected this to be the default behavior. If no `version` is specified, then I would expect that to imply "any version" (which defaults to HEAD when no version was already present locally) rather than "strictly the version that the upstream repo considers default".

> I don't know anything about your use case 

We have fast moving, complex project with a large team of internal and external collaborators where we need to introduce new packages on a regular basis. Rather than having everyone on the team manually clone new packages into their various workspaces whenever they need new things, we provide a tool that they can run to fetch any new packages that may have been introduced to the project.

At first we wrote custom machinery to automatically clone the repos and, when necessary, check out specific versions (because for some repos we do want specific versions while for other actively developed repos it's less important). But then we realized that we could just dump that effort and use vcstool instead. The `.repos` file format was especially nice for organizing and describing what repos were needed.

However, when it comes to the repos that are being actively developed by members of the team, it can be irritating if this tool is automatically changing the branches that they're actively working on. And if there are any uncommitted changes in one of those branches, then the tool errors out because the git checkout command will return a non-zero exit code.

> but why are the branch names in the .repos file and the local working copy different in the first place?

For some repos the `.repos` file does not specify a branch name at all. Those versionless repo entries were meant to suggest that any version is to be accepted for that repo.

> I am not sure the behavior you describe is desired.

I understand that the proposed behavior may be controversial and undesirable for some users, so that's why it's guarded by an optional command line flag. It needs to be explicitly turned on, or else the old behavior will continue to be the default.

dirk-thomas:
> When the state is not described (i.e. no version is specified), then what should be considered the exact described state?

Imo the same behavior as the git CLI is desired. If you `git clone` a repo you will checkout the default branch.

> Rather than having everyone on the team manually clone new packages into their various workspaces whenever they need new things, we provide a tool that they can run to fetch any new packages that may have been introduced to the project.

How about adding an option to `vcs import` instead - something like `--skip-existing` - which simply won't touch directories which already exist? Such an options seems to much more closely describe your intended behavior.

mxgrey:
> How about adding an option to vcs import instead - something like --skip-existing - which simply won't touch directories which already exist?

This would suit my use case. Although I think the most sensible behavior is it should only skip existing directories if they match the specified URL, i.e. [this check](https://github.com/dirk-thomas/vcstool/blob/1f19788136b9af249140c3315458414f52c46bf5/vcstool/clients/git.py#L189-L210) and [this check](https://github.com/dirk-thomas/vcstool/blob/1f19788136b9af249140c3315458414f52c46bf5/vcstool/clients/hg.py#L113-L132) should still be performed.

At that point I think the current PR is exactly what's being prescribed, except the argument `--ignore-head` would be changed to `--skip-existing`.

dirk-thomas:
> Although I think the most sensible behavior is it should only skip existing directories if they match the specified URL ...

I disagree on this. I don't see a different between a user switching to custom branch or a fork - in either of the two cases the new option should skip touching that existing working copy imo. Therefore the new option should skip any existing directory - independent of its state.

mxgrey:
The idea would be to verify that the existing directory is actually a copy of the intended repo and not a random directory that the user created at some time. If it's some random other directory rather than the intended repo, then the tool would exit with an error.

mxgrey:
Also note that the verification looks at whether the intended url is any one of the remotes for the repo, so the local copy can still be primarily aimed at a fork.

dirk-thomas:
> The idea would be to verify that the existing directory is truly a copy of the intended repo and not a random directory that the user created at some time. If it's some random other directory rather than the intended repo, then the tool would exit with an error.

Sounds good to me. Then my previous proposal sounds reasonable:

> Therefore the new option should skip any existing directory - independent of its state.

Depending on the state of the existing directory the tool will either:
* just skip it when the directory contains a working copy which at least has one remote with the specified URL or
* return an error for that directory if all remotes are unrelated or the directory doesn't even contain a working copy at all

Either way the directory is not being modified (e.g. to a different remote / branch).

Would you be able to update the PR with the revised logic?

mxgrey:
That sounds good to me. The only change that needs to be made to this PR then is renaming the `--ignore-head` flag to `--skip-existing`.

I'm traveling without a computer this weekend, but I'll see if GitHub's mobile UI can do the job.

mxgrey:
I've attempted to make the changes with my phone, but I won't be able to test it until I have computer access.

dirk-thomas:
It would be good to also add tests covering the discussed cases.

mxgrey:
I've added tests for each of the conditions that have been discussed, and they seem to be passing.

dirk-thomas:
Thanks for iterating on the patch.

I tried to integrate the new tests into the existing `.repos` file (to also cover the no-version entry in the other commands) and merge the tests into the existing function `test_reimport` since it semantically fits in there.

It took a few more iterations than I expected but the current state looks good. Can you please give it a try with the current patch and comment if it still works for your use case. Thanks.

mxgrey:
Yep, this worked great, thanks!

dirk-thomas:
FYI #136 to also skip non-git existing directories.

PR #100, opened 2019-07-10
export unique tag names when applicable

### LINKED ISSUE #98: Export tags?
I have a bunch of repos that are all checked out to a git tagged release. When I export my workspace, the versions are all recorded as git hashes. I didn't see a way to export the tag instead. I'd like the tagged release so I can more easily understand how packages are being updated. Is this possible?

dirk-thomas:
Closes #98.

### DIFF
--- a/test/export_exact.txt
+++ b/test/export_exact.txt
@@ -6,4 +6,4 @@ repositories:
   tag:
     type: git
     url: https://github.com/dirk-thomas/vcstool.git
-    version: bf9ca56de693a02b93ed423bcef589259d75eb0f
+    version: 0.1.27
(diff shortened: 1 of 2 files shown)


murrdpirate:
Thanks very much for implementing this!

Occasionally, I get a hash instead of the tag. It appears that when the `GitClient` executes the `ls-command`,  it sometimes gets an error like this:

`'fatal: unable to get credential storage lock: File exists', u'4371fd82c77791f1fd1c707aac12e15415728b53\trefs/tags/1.0.0'`

As you can tell from the error, I'm having git store my credentials, but this hasn't led to any other issues in `vcstool`.

It appears that I can resolve this by setting `workers = 1`, so I assume it has something to do with workers competing over the credential lock, but I don't understand why this wouldn't have happened in other situations (e.g. cloning the repos).

PR #104, opened 2019-08-05
import `--force` should overwrite non-empty directories

sloretz:
This makes `import --force` overwrite directories, even if they're not empty and not a git repository. I have only tested this with `git`. I haven't checked the other clients, so I don't know if they're also affected.

### DIFF
--- a/test/test_commands.py
+++ b/test/test_commands.py
@@ -159,6 +159,21 @@ class TestCommands(unittest.TestCase):
             ['git', 'remote', 'remove', 'foo'],
             stderr=subprocess.STDOUT, cwd=cwd_vcstool)
 
+    def test_import_force_non_empty(self):
+        workdir = os.path.join(TEST_WORKSPACE, 'force-non-empty')
+        os.makedirs(os.path.join(workdir, './vcstool/not-a-git-repo'))
+        try:
+            output = run_command(
+                'import', ['--force', '--input', REPOS_FILE, '.'],
+                subfolder='force-non-empty')
+            expected = get_expected_output('import')
+            # newer git versions don't append ... after the commit hash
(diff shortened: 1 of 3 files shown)


test/test_commands.py
```diff
@@ -159,6 +159,21 @@ def test_reimport(self):
             ['git', 'remote', 'remove', 'foo'],
             stderr=subprocess.STDOUT, cwd=cwd_vcstool)
 
+    def test_import_force_non_empty(self):
+        workdir = os.path.join(TEST_WORKSPACE, 'force-non-empty')
+        os.makedirs(os.path.join(workdir, './vcstool/not-a-git-repo'))
```
dirk-thomas: The leading `./` seems unnecessary. Also instead of hard coding the forward slash this could be separate positional arguments: `os.path.join(workdir, 'vcstool', 'not-a-git-repo')`.

dirk-thomas:
Maybe the [help text should](https://github.com/dirk-thomas/vcstool/blob/df7da86566f66e0e57899ce9ad09fcf3d1dc0a8e/vcstool/commands/import_.py#L42-L43) be updated to clarify that any arbitrary content will be removed, e.g.:

> Delete existing directories if they don't contain the same repositories

test/test_commands.py
sloretz: Directories are `os.path.join()`'d in 53b4eb71ec44bd204458187de948649245ae1bd1

sloretz:
> Maybe the help text should be updated to clarify that any arbitrary content will be removed, e.g.:

@dirk-thomas updated help text in 83da17a

dirk-thomas:
Thanks for the patch!

Issue #109, opened 2019-12-02
vcs export --exact reports tag names instead of commit hash


crdelsey:
**Version:** vcs 0.2.4

**How to Reproduce:**
Run `vcs export --exact` on a set of repos.

**Expected behavior:**
The version field of each repo will report the exact commit hash like this:
```
  src/ros2/rclpy:
    type: git
    url: https://github.com/ros2/rclpy.git
    version: 76d26a125da7f9df19ba35beb60cb16838019a10
```

**Actual Behavior:**
Frequently the hash is replaced by a tag name.
```
  src/ros2/rcl_interfaces:
    type: git
    url: https://github.com/ros2/rcl_interfaces.git
    version: 0.8.0
```

---

I think when the --exact parameter is specified, it is preferable to provide the commit hash.

dirk-thomas:
> I think when the --exact parameter is specified, it is preferable to provide the commit hash.

The behavior to return tags when available is intentional (see #98). Since tags are intended to be immutable and significantly more readable they are being used by `--exact` if the current hash has exactly one tag pointing to it.

crdelsey:
I don't know about all vcs tools, but tags frequently aren't all that immutable. For instance, they are frequently used to mark the `latest_stable` version or some such thing, with the tag being moved as someone or CI decides that a new version is now passing the tests.

I'm pretty sure they can be moved in Git, so, for right or wrong, if somebody moves the tag, the --exact option is not so exact anymore.

Having the tag is a nice feature most of the time, but, it disagrees with the docs for the --exact parameter.
```
"export" command parameters:
  --exact            Export exact commit hash instead of branch names
```

dirk-thomas:
> I'm pretty sure they can be moved in Git, so, for right or wrong, if somebody moves the tag, the --exact option is not so exact anymore.

Sure, you can also rewrite history in git - still something to consider not doing ;-)

> Having the tag is a nice feature most of the time, but, it disagrees with the docs for the --exact parameter.

Agreed, clarifying the help text would certainly be good :+1:

> If you want to keep this behavior, maybe another option, something like --hash-only could be used to disable the tag lookup feature?

Adding another option to toggle the behavior to always return hashes vs. potentially tags if they are unique would be good. Since the change has only landed recently I am not too concerned about changing the behavior (if we think that makes more sense). I am just not sure what the right names for the two options would be `--exact` and ? `--hash-only` seems very git specific...

dirk-thomas:
[...]

PR #115, opened 2020-01-03
Add --exact-with-tags option to export command.

crdelsey:
This addresses #109.

This adds a new command line option to the `export` sub-command: `--exact-with-tags`.

`--exact-with-tags` does what `--exact` does currently. That is, it reports the exact hash ID of the current commit or replaces it with a tag name if one can be found.

`--exact` now only reports the exact hash ID and does not replace with a tag.

`--exact` is a bit friendlier to automation, while `--exact-with-tags` is a bit friendlier to humans

Note that the "with-tags" part is ignored by all version control tools except Git.


crdelsey:
I want to do a bit of refactoring on the export method still. That's why this is a draft PR.

dirk-thomas:
I [rephrased the help messages](https://github.com/dirk-thomas/vcstool/pull/115/commits/6b96eca770c5e6ce88db1df9a8b6029bb39b973a) a bit and [de-duplicated the return logic](https://github.com/dirk-thomas/vcstool/pull/115/commits/5e9944a002bc92a9c5851fdb06db9e3c291b102f). I hope that makes sense to you.

While my changes also address some linter failures the tests checking the `export` command still fail. Please update them to pass again as well as extend them to cover the new option.

> I want to do a bit of refactoring on the export method still.

Please keep any independent refactoring in a separate PR so this has an as minimal as possible diff.

crdelsey:
> Please keep any independent refactoring in a separate PR so this has an as minimal as possible diff.

Ok. I'll defer the refactoring. You took care of the thing that was bothering me most anyhow; the duplication of the return logic. Thanks for that, by the way.

The CI failure seems unrelated to this change, so I'm going to go ahead and mark this PR ready.

dirk-thomas:
> The CI failure seems unrelated to this change, so I'm going to go ahead and mark this PR ready.

I addressed the CI failure in #117. Please rebase your branch on top of the latest state of the master branch to pass CI.

crdelsey:
Thanks. That did it and it is passing CI now.

dirk-thomas:
@crdelsey Thanks for the contribution.

@murrdpirate FYI since you requested tags being used in the exact export and after this you will need to use the new option `--exact-with-tags`.

murrdpirate:
Thanks for the heads up!

JanosToth1989:
Hi,
I'm using the tool to create release from more repositories on github. I tag my current version and run the exact-with tags but always get the commit hash like exact. How can I get tags in the yaml file? 
I installed vcs via ROS2 Humble.

PR #116, opened 2020-01-04
Fix output on failed checkout

chapulina:
Without this fix, a failed checkout will print a vcstool internal error:

```
=== src/ign-common (hg) ===
Invocation of command 'import' on client 'hg' failed: UnboundLocalError: local variable 'result_clone' refer
enced before assignment (/usr/local/lib/python3.6/dist-packages/vcstool/clients/hg.py:169) 
```

With the fix, we get the desired message, in my case:

```
=== src/ign-common (hg) ===
Could not checkout 'custom_paths_uri': abort: uncommitted changes
(commit or update --clean to discard changes)
```


dirk-thomas:
Thanks for the bug fix.

PR #122, opened 2020-02-27
Fix encoding issue when running command on non-UTF8 system

delijo:
On Windows platform configured with non-english locale, the date displayed by `svn info` is in the locale format even if the rest of the svn output is in english.
The date can then contain accentuated characters or special punctuation and doesn't convert well since the encoding is not UTF-8 on this platform.

With this fix, *vcstool* won't assume the process will be run with UTF-8 encoding but let Python libs determine the right encoding.
`universal_newlines` argument is used instead of `text` to support more Python versions (`text` requires Python 3.7).

### DIFF
--- a/vcstool/clients/vcs_base.py
+++ b/vcstool/clients/vcs_base.py
@@ -79,12 +79,12 @@ def run_command(cmd, cwd, env=None):
     try:
         proc = subprocess.Popen(
             cmd, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
-            env=env)
+            env=env, universal_newlines=True)
         output, _ = proc.communicate()
-        result['output'] = output.rstrip().decode('utf8')
+        result['output'] = output.rstrip()
         result['returncode'] = proc.returncode
     except subprocess.CalledProcessError as e:
-        result['output'] = e.output.decode('utf8')
+        result['output'] = e.output
         result['returncode'] = e.returncode
     return result
 


delijo:
@dirk-thomas : Hi, any plan to integrate this and do a new release ?

dirk-thomas:
I am sorry for the delay. This fell off my radar.

I am not sure this patch will be ok since in Python 2 the default encoding is `ASCII`. So I am hesitant to merge this as is.

delijo:
I tested locally with Python 2.7.18rc1 and it is working fine (Windows with fr_FR locale).
Not sure if we should worry more about Python 2 since it is deprecated from a long time and as officially reached [EOL on January 1, 2020](https://www.python.org/doc/sunset-python-2/) (even if the very-final release will come in 3 days).
People still using Python 2 can install Python 3 in parallel to run this script as an easy workaround. More easy, in my opinion, than patching the file manually like me and my colleagues are doing to have something working.

delijo:
Work also for me on Ubuntu Linux with Python 2.7.17 with fr_FR locale.
On my Ubuntu, `locale.getpreferredencoding(False)` returns _ANSI_X3.4-1968_.
On my Windows, `locale.getpreferredencoding(False)` returns _cp1252_.

dirk-thomas:
> Not sure if we should worry more about Python 2 since it is deprecated from a long time and as officially reached EOL on January 1, 2020 (even if the very-final release will come in 3 days).

All active ROS 1 distros (Kinetic and Melodic) are using Python 2 and will continue to do so until their EOL. So yes, we have to still make sure that things continue to work with Python 2 (assuming new version of Python packages like this are continued to be released into Ubuntu distros like Xenial and Bionic).

> People still using Python 2 can install Python 3 in parallel to run this script as an easy workaround.

It must still be ensured that existing users having the Python 2 package installed and updating to the latest version are not getting a regression.

Regarding the current patch: using `universal_newlines` leaves the decoding to Python with no option to configure how `UnicodeDecodeError`s are handled. How about an the following:
* don't use `universal_newlines` (as it was before)
* readd the `decode()` calls (as before)
* but instead of passing `utf-8` explicitly relying on Python to pick the default encoding
* and to make this robust to cases which the decoding would fail pass `errors='replace'`

To summarize - before:

> `output.rstrip().decode('utf8')`

after:

> `output.rstrip().decode(errors='replace')`

I tried this with various inputs and it worked for me but please try it with your specific locale and content where you currently encounter the issue with the unpatched version.

If that doesn't work maybe try:

> `output.rstrip().decode(locale.getpreferredencoding(False), errors='replace')`

dirk-thomas:
@delijo Friendly ping.

dirk-thomas:
@delijo Friendly ping.

delijo:
Hum, I reverted my local changes and I am not able to reproduce the issue so it is not a good start to test your way of fixing it.
I don't know what has changed since my reporting. I will need to investigate more.

delijo:
Hi,
I encountered a similar issue today (probably not the same cause by causing UnicodeDecodeError at the same place) and tested your solution.
It almost work: there isn't a UnicodeDecodeError with your solution but the text displayed on the console is weird because it shows strange characters due to replace strategy.
With my solution, the good characters are printed on the console.

dirk-thomas:
> I encountered a similar issue today (probably not the same cause but causing UnicodeDecodeError at the same place) and tested your solution.
It almost work: there isn't a UnicodeDecodeError with your solution but the text displayed on the console is weird because it shows strange characters due to replace strategy.

Please share the exact steps / environment to reproduce the similar issue.

delijo:
The issue was during execution of `vcs` on a French Windows 10 installation: the destination folder was existing causing the `svn` cli to display an error message with OS error string in French so containing accentuated characters using codepage 1252 so not converting well using UTF8 decode.

dirk-thomas:
> Please share the exact steps / environment to reproduce the similar issue.

E.g. a Dockerfile.

dirk-thomas:
@delijo Friendly ping.

dirk-thomas:
@delijo Friendly ping.

dirk-thomas:
Closing due to no further replies.

PR #123, opened 2020-04-17
Allow fetching repos from url

rotu:
If a URL is passed to --input, we fetch the resource. This removes the need for curl or wget in many cases.

### DIFF
--- a/vcstool/commands/import_.py
+++ b/vcstool/commands/import_.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 import argparse
 import os
 import sys
+import urllib.request
 
 from vcstool.clients import vcstool_clients
 from vcstool.clients.vcs_base import run_command
@@ -33,12 +34,20 @@ class ImportCommand(Command):
         self.recursive = recursive
 
 
+def file_or_url_type(x):
+    try:
+        return urllib.request.urlopen(x)
+    except ValueError:
+        return argparse.FileType('r')(x)
+
+
 def get_parser():
     parser = argparse.ArgumentParser(
         description='Import the list of repositories', prog='vcs import')
     group = parser.add_argument_group('"import" command parameters')
     group.add_argument(
-        '--input', type=argparse.FileType('r'), default=sys.stdin)
+        '--input', type=file_or_url_type, default='-',
+        help="Where to read YAML from.", metavar='FILE_OR_URL')
     group.add_argument(
         '--force', action='store_true', default=False,
         help="Delete existing directories if they don't contain the "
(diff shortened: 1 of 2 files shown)


vcstool/commands/import_.py
```diff
@@ -33,12 +38,20 @@ def __init__(self, args, url, version=None, recursive=False):
         self.recursive = recursive
 
 
+def file_or_url_type(x):
+    try:
+        return urlopen(x)
```
dirk-thomas, line 43: In case the URL parsing raises an exception a stacktrace is shown which isn't user friendly. Please catch the relevant exceptions and print a single line error message.

vcstool/commands/import_.py
```diff
@@ -33,12 +38,20 @@ def __init__(self, args, url, version=None, recursive=False):
...
+        return argparse.FileType('r')(x)
+
+
 def get_parser():
     parser = argparse.ArgumentParser(
         description='Import the list of repositories', prog='vcs import')
     group = parser.add_argument_group('"import" command parameters')
     group.add_argument(
-        '--input', type=argparse.FileType('r'), default=sys.stdin)
+        '--input', type=file_or_url_type, default='-',
```
dirk-thomas, line 53: The loading of the URL should happen and arg parsing since it could take an arbitrary amount of time. Considering the case that e.g. completion is being computed using `parse_known_args`, Please move it loading of the URL post arg parsing.

dirk-thomas: Unrelated question: why does the patch change the `default` argument?

rotu:
> Beside the inline comments please also add tests covering the new feature.

You have previously objected when I've added tests and features in the same PR. Would you still like me to add tests?

vcstool/commands/import_.py
rotu: Because argparse formats it weird otherwise. `-` is another name for stdin.

dirk-thomas:
> You have previously objected when I've added tests and features in the same PR.

I doubt that I have ever made such a statement - to not add tests for the feature being added in the PR.

> Would you still like me to add tests?

Yes.

dirk-thomas:
@rotu Since it has been a month I just wanted to check if you are planning to follow up on this ticket or if not close the ticket for now.

rotu:
It's still a draft since the test coverage is not comprehensive and not a high priority right now. Due to personal stuff, I'm not sure if I'll be able to wrap this up any time soon. I'm sorry - I really want to give this the attention it deserves. If you want to pick up where I left off, I wouldn't mind.

dirk-thomas:
Superseded by #155.

PR #128, opened 2020-04-22
Support authentication via netrc

kyrofa:
`vcs_base.load_url()` currently doesn't support authentication. Add support for both basic and token-based authentication by parsing netrc files. Note that this borrows heavily from vcstools.


kyrofa:
Huh. I'm not sure what I'm doing wrong on the mock bits for earlier pythons.

kyrofa:
Ah ha, it's an [upstream python bug](https://bugs.python.org/issue28380). All fixed now.

dirk-thomas:
> Note that this borrows heavily from vcstools.

As discussed offline it needs to be double checked that this contribution complies with the license.

kyrofa:
Alright, this has been completely rewritten from scratch

vcstool/clients/vcs_base.py
```diff
@@ -93,7 +113,12 @@ def load_url(url, retry=2, retry_period=1, timeout=10):
     try:
         fh = urlopen(url, timeout=timeout)
     except HTTPError as e:
-        if e.code == 503 and retry:
+        if e.code in (401, 404):
+            # Try again, but with authentication
+            fh = _authenticated_urlopen(url, timeout=timeout)
```
dirk-thomas, line 118: Does the sibling function `test_url` not need the same logic?

vcstool/clients/vcs_base.py
```diff
@@ -93,7 +113,12 @@ def load_url(url, retry=2, retry_period=1, timeout=10):
...
+            if fh is not None:
+                return fh.read()
```
dirk-thomas, line 120: Instead of duplicating the return statement how about changing the remaining of this block to only raise in the `else` case?

vcstool/clients/vcs_base.py
```diff
@@ -93,7 +113,12 @@ def load_url(url, retry=2, retry_period=1, timeout=10):
...
+        elif e.code == 503 and retry:
```
dirk-thomas, line 121: In the case of a retry the next iteration would again first try a `urlopen` just to then use `_authenticated_urlopen` again. It would be better if a retry will immediately use the authenticated version if it was necessary in the first iteration.

vcstool/clients/vcs_base.py
```diff
@@ -132,3 +157,94 @@ def test_url(url, retry=2, retry_period=1, timeout=10):
...
+
+
+def _authenticated_urlopen(uri, timeout=None):
+    machine = urlparse(uri).netloc
+    if not machine:
+        return None
+
+    credentials = _credentials_for_machine(machine)
+    if credentials is None:
+        logger.warning('No credentials found for "%s"' % machine)
```
dirk-thomas, line 169: I guess this poorly interacts with the progress dots, no? It also wouldn't appear in a repository specific output section which makes it difficult to figure out which repo is causing this. Maybe reporting this error case somehow differently would be better...

.travis.yml
```diff
@@ -12,7 +12,7 @@ matrix:
 install:
   # newer versions of PyYAML dropped support for Python 3.4
   - if [ $TRAVIS_PYTHON_VERSION == "3.4" ]; then pip install PyYAML==5.2; fi
-  - pip install coverage flake8 flake8-docstrings flake8-import-order pytest PyYAML
+  - pip install appdirs coverage flake8 flake8-docstrings flake8-import-order pytest PyYAML mock
```
dirk-thomas, line 15: The dependency on `appdirs` needs to be declared in the `setup.py` as well as `stdeb.cfg` file.

vcstool/clients/vcs_base.py
kyrofa: Actually if it hits this branch of code it never hit `_authenticated_urlopen`, but your point is still valid: `_authenticated_urlopen()` doesn't support the retry logic, and it should indeed. I'll add it.

vcstool/clients/vcs_base.py
kyrofa: Yeah it's not ideal, but I couldn't immediately think of a better way to do it. Shall I remove it for now? It will be fairly clear that authentication doesn't work because the pull will fail, so I'm not entirely sure how helpful this is.

.travis.yml
kyrofa: Whoops, thank you.

vcstool/clients/vcs_base.py
kyrofa: Good catch, I missed that feature entirely :blush: .

vcstool/clients/vcs_base.py
kyrofa: Good idea. Refactored in 1cdc397997a4c9abd38717fb7cc08a4ada95d58a.

.travis.yml
kyrofa: Fixed in 92a3bdf8e4183469bad475e67a0a1af4d8695855.

vcstool/clients/vcs_base.py
kyrofa: I've gone ahead and removed it in b49456971293af1141f9df1d0f051a32b4107fec.

vcstool/clients/vcs_base.py
kyrofa: Fixed in b49456971293af1141f9df1d0f051a32b4107fec.

vcstool/clients/vcs_base.py
kyrofa: Added in f3286273b5150016e991a51173e3ab6cfa2e13fd.

kyrofa:
> Please add some documentation about the new auth files to the README.

Done in fc361a2287a953e9a274d1b2d37fa6fc3360ad8e, does that seem to be a sensible place to put it?

README.rst
```diff
@@ -138,6 +138,44 @@ The set of repositories to operate on can optionally be restricted by the type:
 If the command should work on multiple repositories make sure to pass only generic arguments which work for all of these repository types.
 
 
+Access repositories that require authentication
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Vcstool supports authenticated access to repositories by parsing files with a netrc-like format.
```
dirk-thomas, line 144: I think this should mention that this is only relevant for the `tar` and `zip` types (assuming my assumption is right).

vcstool/clients/vcs_base.py
```diff
@@ -115,20 +132,145 @@ def test_url(url, retry=2, retry_period=1, timeout=10):
...
+
+        return _retryable_function
+    return _retry_decorator
+
+
+def _authenticated_urlopen(uri, retry=2, retry_period=1, timeout=10):
+    if isinstance(uri, str):
+        request = Request(uri)
+    else:
+        request = uri
```
dirk-thomas, line 181: I am wondering if this might be an issue if the passed `uri` is of type `unicode` in Python 2. Maybe the logic should be inverted to:

```
if isinstance(uri, Request):
    request = uri
else:
    request = Request(uri)
```

vcstool/clients/vcs_base.py
```diff
@@ -115,20 +132,145 @@ def test_url(url, retry=2, retry_period=1, timeout=10):
...
+    try:
+        credentials = netrc.netrc(file=filename).authenticators(machine)
+    except EnvironmentError as e:
+        # Don't error just because the file didn't exist or we didn't have
+        # permission to access it. Catching this situation this way to be
+        # compatible with python 2 and 3.
+        if e.errno not in (errno.ENOENT, errno.EACCES):
+            raise
+    except netrc.NetrcParseError:
+        # If this file had issues, don't error out so we can try fallbacks
```
dirk-thomas, line 273: Would a warning be more helpful to the user rather then silently ignoring it? On the other hand this behavior might match how other programs using `.netrc` work.

vcstool/clients/vcs_base.py
kyrofa: Ah, good catch, this would be `basestring` in python2. I agree, your approach should work better for both. Fixed.

README.rst
kyrofa: Your assumption is correct, rephrased.

vcstool/clients/vcs_base.py
kyrofa: We're a little bit in uncharted territory here, to be honest. My model for the design is `apt`, but one generally runs that as root, so running into permissions issues accessing these files isn't normally an issue. It could be here, if the machine admin only wanted members of a particular group to be able to fetch repositories using these credentials. I had mixed feelings about warning about this or quietly proceeding, so I'm happy adding a warning if you like.

kyrofa:
Gentle ping here.

kyrofa:
Hey @dirk-thomas, how are you feeling about this?

0xjairo:
@kyrofa  / @dirk-thomas  is there a reason why this MR was not accepted?

PR #129, opened 2020-04-25
Fix crash under Windows when files are readonly

### LINKED ISSUE #130: import --force is broken on Windows
`vcs import --force` fails with a PermissionError on Windows.
This is because git creates its index files with the readonly flag, which must be cleared before the file can be deleted.

```Invocation of command 'import' on client 'git' failed: PermissionError: [WinError 5] Access is denied: 'src/eProsima/Fast-CDR' (c:\python37\lib\site-packages\vcstool\clients\git.py:242)```

https://github.com/ros2/rmw_cyclonedds/pull/168/checks?check_run_id=627448665#step:4:430

### PR DESCRIPTION
**Pull request by rotu**
This frequently happens when overwriting a git repository with `vcs import --force`

Fix #130

### DIFF
--- a/vcstool/clients/vcs_base.py
+++ b/vcstool/clients/vcs_base.py
@@ -79,11 +79,12 @@ class VcsClientBase(object):
             try:
                 os.remove(f)
             except OSError as e:
-                if getattr(e, 'winerror') != 5:
+                if getattr(e, 'winerror', None) == 5:
+                    # on Windows you need to clear the readonly bit first
+                    os.chmod(f, stat.S_IWRITE)
+                    os.remove(f)
+                else:
                     raise
-                # on Windows you need to clear the readonly bit first
-                os.chmod(f, stat.S_IWRITE)
-                os.remove(f)
 
         # If path is a symlink or a file, delete it.
         try:
--- a/vcstool/clients/vcs_base.py
+++ b/vcstool/clients/vcs_base.py
@@ -128,6 +128,7 @@ class VcsClientBase(object):
                             "Could not remove directory '%s': %s" % (path, e),
                         'returncode': 1
                     }
+        return None
 
 
 def run_command(cmd, cwd, env=None):
--- a/vcstool/clients/vcs_base.py
+++ b/vcstool/clients/vcs_base.py
@@ -1,3 +1,4 @@
+import errno
 import os
 import socket
 import stat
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by rotu** [comment]
This factored out the logic so that I would not need to repeat the fix in many places. It will remain one PR as this is an implementation detail of the proposed fix, not an orthogonal change.

Please review this PR as-is.

File: vcstool/clients/vcs_base.py [hunk 1]
```diff
@@ -71,6 +74,64 @@ def _create_path(self):
...
         return None
 
+    def _create_or_truncate_path(self):
+        def remove_file(f):
+            try:
+                os.remove(f)
+            except OSError as e:
+                if getattr(e, 'winerror', None) == 5:
+                    # on Windows you need to clear the readonly bit first
+                    os.chmod(f, stat.S_IWRITE)
```
dirk-thomas, line 84: I am hesitant to unconditionally change file attributes. The scope of this should be narrowed down to the specific case where it is desired (the `.git` directory).

dirk-thomas:
> Please review this PR as-is.

I have done so and given you my feedback. Please consider following it (or not).

vcstool/clients/vcs_base.py
rotu: If other files have the `readonly` attribute, why should a `import --force` fail instead of deleting the file? I think it's more expected that, like with Linux, if the file can be deleted, it will be.

dirk-thomas:
Please also add a unit test for the use case (and rebase your branch).

rotu:
Backmerged master. Feel free to contribute, in a separate PR, a unit test to make sure `--force` works as documented.

dirk-thomas:
> Feel free to contribute, in a separate PR, a unit test to make sure --force works as documented.

The test should not be separate from the actual patch since it is supposed to ensure that the patch does achieve what it indents to.

I am sorry to hear that you don't want to iterate on this PR (on the previous feedback as well as the unit test). I will go ahead and close this ticket and work on a patch myself.

rotu:
> I am sorry to hear that you don't want to iterate on this PR (on the previous feedback as well as the unit test). I will go ahead and close this ticket and work on a patch myself.

Awesome! [Cunningham's law](https://en.wikipedia.org/wiki/Ward_Cunningham#Cunningham%27s_Law) at work! Thank you for being a responsible steward of your own project.

dirk-thomas:
> Cunningham's Law states "the best way to get the right answer on the internet is not to ask a question; it's to post the wrong answer."

Unfortunately I have to disagree with you on the opinion that this pull request and it's iteration is anywhere close to "the best way".

PR #131, opened 2020-05-05
add GitHub action workflow

dirk-thomas:
Using GitHub action workflow to run CI on:
* Ubuntu with all targeted Python versions
  * except 3.4 which is not supported by GitHub action, keeping Travis CI around while Debian Jessie with Python 3.4 is targeted
* macOS with the latest Python release (atm 3.8)
* Windows with the latest Python release (atm 3.8)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+name: vcstool
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows
```diff
@@ -0,0 +1,25 @@
+name: vcstool
```
christophebedard, line 1: just in case you're wondering why the workflow isn't getting executed, this file has to be a `*.yml` file in the `.github/workflows/` directory

dirk-thomas: Thanks. Good point. First try :wink:

PR #134, opened 2020-05-07
fix issues on Windows

### LINKED ISSUE #130: import --force is broken on Windows

dirk-thomas:
The first commit fixes #130 and fixes ros2/ci#200. It makes sure that files/directories which are hidden can be deleted by non-admin users. This replaces #129.

The second commit handles a problem in the tests which determines the git version. On Windows the reported string might contain a version component "windows".

### DIFF
--- a/test/test_commands.py
+++ b/test/test_commands.py
@@ -1,9 +1,11 @@
 import os
-import shutil
 import subprocess
 import sys
 import unittest
 
+sys.path.insert(0, os.path.dirname(os.path.dirname(__file__)))
+
+from vcstool.util import rmtree  # noqa: E402
 
 REPOS_FILE = os.path.join(os.path.dirname(__file__), 'list.repos')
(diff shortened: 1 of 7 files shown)


dirk-thomas:
While the tests pass with the GitHub actions on Windows this change was also tested manually on Windows without admin privileges. Without this change the tests can't be run repeatedly since after the first run the temporary test workspace can't be deleted (due to the issue described in #130).

rotu:
Did you mean to split this into two PRs to keep unrelated changes separate?

PR #143, opened 2020-05-21
add double dash to move to another branch, not reset file/directory in git client

firemark:
This bug occurs only when you have file and branch on remote with the same name, because if `git checkout` don't find local branch then resets file with this name.

### How to reproduce
1. Make a file `foo` and commit it
2. Make a branch `foo` and push it to remote
3. Use vcs tool with config:
```
repositories:
  repo:
    type: git
    url: git@gitlab.com:foo/bar/repo.git
    version: foo
```
4. In results, downloaded repository has branch `master` not `foo` 

### Solution
According to https://git-scm.com/docs/git-checkout adds a double dash is enough

PS. thanks to @ksuszka to found the bug!

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -320,7 +320,7 @@ class GitClient(VcsClientBase):
 
         if checkout_version:
             cmd_checkout = [
-                GitClient._executable, 'checkout', checkout_version]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch.

dirk-thomas:
See #164 which reports that this breaks with older git versions.

PR #147, opened 2020-06-10
 add --shallow option to import verb, git-only 

dirk-thomas:


ipa-mdl:
This fails, if version is `HEAD`. Not sure of this is a supported use case.
I have been using this to select the default branch explicitly.

dirk-thomas:
To cite the docs:

> The HEAD in Git is the pointer to the current branch reference, which is in turn a pointer to the last commit you made or the last commit that was checked out into your working directory.

So without a working directory the term `HEAD` doesn't seem to make much sense.

> I have been using this to select the default branch explicitly.

Can you please try it without a version instead?

ipa-mdl:
> So without a working directory the term HEAD doesn't seem to make much sense.

Yes, I know. That's why I wrote: "Not sure of this is a supported use case."
It made sense (beforehand), because the version was changed after cloning the repository.

> Can you please try it without a version instead?

I works (and worked).
But IMHO this is not expressive enough, at least not in my use case.
I think I will keep the support for `HEAD` on my end and just drop the version in this case.
`HEAD` is a reserved name and cannot collide with any other branch or tag :)

dirk-thomas:
> But IMHO this is not expressive enough, at least not in my use case.

Can you elaborate how it is "not expressive enough"? It seems to be the same if you you `git clone URL` without passing a specific branch name.

dirk-thomas:
I just added a note to the README mentioning the default value of the `version` key which wasn't documented before: #150.

PR #153, opened 2020-06-15
Fix text-related typos/mistakes and improve consistency

christophebedard:
Some minor text-related fixes.


dirk-thomas:
Thank you for the improvements.

Issue #103, opened 2019-08-02
Offline tests


jayvdb:
Would be possible to have a cut down version of the test suite which uses offline data.

I'm packaging this for openSUSE https://build.opensuse.org/package/show/home:jayvdb:ros/python-vcstool , and the rpm build environment is isolated from the internet.

I can add tarballs of other repos from the tests repo.list to the build environment, and manually import them.

Ideally the test runner would recognise that the repo is already imported, and not try to update it, and then run the tests.

dirk-thomas:
I am not sure how valuable a cut down version of the test suite would be? I would guess a reduced set will inevitably reduce coverage.

Please feel free to propose patches which would help with your use case. The only precondition I would have is that it shouldn't negatively affect the tests when run through the current CI as well as locally for a developer.

dirk-thomas:
@jayvdb Are you interested to work on this? If yes, I am happy to keep this ticket open.

jayvdb:
My plate is extremely full, especially at the moment due to GSoC duties, so I can not commit to doing it in the near future, but it does need to get done for distro packaging.  It is also blocking packaging of the tools which use vcstool as a dependency, generally the "future" ROS toolkit being developed.  As this package rises in importance, the need for this will increase, and people willing and able to do it also increases.

The proposed changes would not alter the existing CI.  They would only add support for the existing test suite to work correctly in distro packaging isolated VM environments, provided a limited smoke test.  Without it, distro packages bitrot because the *packaging* is not maintained, e.g. dependency changes are not noticed during version bumps.

dirk-thomas:
@jayvdb Friendly ping.

rotu:
Agreed. I'm trying to make tests for https://github.com/dirk-thomas/vcstool/pull/123 without adding remote web requests. I think this can be done by simply spinning up a `http.server.HTTPServer` on localhost in a parallel thread.

dirk-thomas:
I am going to close this ticket since I don't plan to spend any time on this.

If someone would like to contribute a pull request I am more than happy to review it.

PR #155, opened 2020-06-27
Allow fetching repos file from URL

christophebedard:
Supersedes and addresses review feedback from #123

I added a simple test, but the actual request could probably easily be mocked (#103).


vcstool/commands/import_.py
```diff
@@ -36,12 +41,28 @@ def __init__(
         self.shallow = shallow
 
 
+def file_or_url_type(x):
+    if x.startswith(('http://', 'https://')):
```
dirk-thomas, line 45: Instead of explicitly whitelisting schemes how about inverting the logic and check if the passed value is a valid file path?

vcstool/commands/import_.py
```diff
@@ -36,12 +41,28 @@ def __init__(
...
+        try:
+            # use another user agent to avoid getting a 403
```
dirk-thomas, line 47: Can you please elaborate when you expect to get a 403 without the `User-Agent` header.

vcstool/commands/import_.py
```diff
@@ -190,8 +211,11 @@ def main(args=None, stdout=None, stderr=None):
         path_help='Base path to clone repositories to')
     args = parser.parse_args(args)
     try:
-        repos = get_repositories(args.input)
-    except RuntimeError as e:
+        input_ = args.input
+        if isinstance(input_, request.Request):
+            input_ = request.urlopen(input_)
+        repos = get_repositories(input_)
+    except Exception as e:
```
dirk-thomas, line 218: This hides the stacktrace for all exception types - even unexpected ones. Therefore this should only catch specific / expected exception here.

vcstool/commands/import_.py
christophebedard: by "valid" do you mean "existing"?

If so, then if the file does not exist, it will try using the path as a URL, and it will print an HTTP/URL-related error, i.e. it will never say that the file just doesn't exist.

dirk-thomas: > by "valid" do you mean "exists"?

Yes, that was my idea.

> If so, then if the file does not exist, it will try using the path as a URL, and it will print an HTTP/URL-related error because it's not really a URL, i.e. it will never say that the file just doesn't exist.

Maybe before trying to load the request it can try to parse the value first and check that it has a scheme part?

vcstool/commands/import_.py
christophebedard: done in 29cc83e5badd0dabbee17d6a3aebb9e8f61246bf

vcstool/commands/import_.py
christophebedard: done in b4daa32222d568748a6f090127338b1543f71dd9.

As an example, I get a 403 if I use this URL without that non-default user agent: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/-/raw/master/instrumented.repos

vcstool/commands/import_.py
christophebedard: see fef2e82d3f3353c2c7fccd04ef4bdfdf5a4ca771, I think it's a good middle ground

vcstool/commands/import_.py
dirk-thomas: Interesting, I didn't know that about GitLab. Can you update the `User-Agent` to instead of pretending to be `wget` state the actual tool. From briefly reading the docs `vcstool/<version>` sounds like an appropriate user agent string (and it seems to work).

christophebedard: done in a5d4092eb667ab28d021246b72381a6cdedfa6d9

I wasn't expecting it to accept a "random"/unknown user agent.

dirk-thomas:
Thanks for the updates. I made two minor changes:
* moved the new function around and renamed the function argument from `x` to `value` and
* reverted the unrelated default value change (I will create a separate PR for that which addresses the same in another place too)

christophebedard:
looks good!

PR #167, opened 2020-07-31
show different symbol for repos which print a warning in yellow

dirk-thomas:
Follow up of #166.


dirk-thomas:
Unfortunately this isn't a proper way to identify repos which should be highlighted. Back to the drawing board...

PR #168, opened 2020-07-31
fix import with git older than 1.8.4.3

### LINKED ISSUE #164: Importing non-local branches with older git version
Running `vcs import`, when a repository specified by a `vcs.yml` already exists, and specifying a branch name in the `version` field results in the following error (assuming the attempted branch checkout is named `test`):

`fatal: invalid reference: test`

This happens when the branch was not previously checked-out locally. The command that is executed by `vcstool/clients/git.py` is: `git checkout test --` which fails also when executed manually. However, executing the following command succeeds: `git checkout test`.

The import command should support checking out remote branches if not available locally so that switching between branches does not fail with the error mentioned.

dirk-thomas:
Fixes #164.

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -1,4 +1,5 @@
 import os
+import subprocess
 
 from vcstool.executor import USE_COLOR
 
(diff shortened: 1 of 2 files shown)


vcstool/clients/git.py
```diff
@@ -413,6 +426,10 @@ def import_(self, command):
             cmd_checkout = [
                 GitClient._executable, 'checkout', checkout_version, '--']
             result_checkout = self._run_command(cmd_checkout)
+            if result_checkout['returncode']:
+                if self.get_git_version() < (1, 8, 4, 3):
```
avishefi, line 430: Throws:
`Invocation of command 'import' on client 'git' failed: TypeError: '<' not supported between instances of 'list' and 'tuple' (/usr/local/lib/python3.6/site-packages/vcstool/clients/git.py:430)`

```suggestion
                if self.get_git_version() < [1, 8, 4, 3]:
```

dirk-thomas: Thanks for the fix. Did this make it work for you with an older version of git?

avishefi: Works great, thank you for the prompt response.

dirk-thomas: Thanks for the feedback and testing.

Issue #170, opened 2020-09-08
vcs is imprecise about git ref types


rotu:
vcs tries to be "smart" about git ref types in a way that is error-prone and limiting.

1. `version` can specify a branch, tag, or hash. But if this is not unique, but vcs provides no way to disambiguate, instead throwing an error when a branch and tag have the same name.
```
vcs import src < repos.yml
=== src/test-repo (git) ===
The version ref is a branch as well as tag but with different hashes
```
and warning when `version` could be either a tag or a hash
```
vcs import src < repos.yml
=== src/test-repo (git) ===
Cloning into '.'...
warning: refname '0246' is ambiguous.
Note: switching to '0246'.
...
```

2. If `version` is neither a branch nor a tag, `vcs` assumes it is a hash, leading to confusing error messages:
```
echo 'repositories: {vcstool: {type: git, url: "https://github.com/dirk-thomas/vcstool.git", version: tags/0.0.0}}' | vcs import src
=== src/vcstool (git) ===
Could not checkout ref 'tags/0.0.0': fatal: invalid reference: tags/0.0.0
```

3. There is not a way to tell vcstool to use a particular ref that is not a hash, branch, or tag:

```
echo 'repositories: {vcstool: {type: git, url: "https://github.com/dirk-thomas/vcstool.git", version: pull/100/head}}' | vcs import
=== ./vcstool (git) ===
Could not checkout ref 'pull/100/head': fatal: invalid reference: pull/100/head
```

vcs should have a way to specify an upstream ref explicitly and flexibly.

dirk-thomas:
> vcs is sloppy about git ref types
> vcs tries to be "smart" about git ref types in a way that is error-prone and limiting.

If you don't want to alienate readers which you would like to help you on your issues it might be helpful to use better language than words like "sloppy".

> It would be nice to just put in the qualified ref (e.g. `heads/1.0.0` vs `tags/1.0.0`). It also doesn't allow refs that are neither a branch nor a tag, e.g. `refs/pull/123/merge`.

Can you please be more specific where you think this more specific refs should be used.

rotu:
> If you don't want to alienate readers which you would like to help you on your issues it might be helpful to use better language than words like "sloppy".

Fixed. I was trying to not state the issue too narrowly. Basically the "version" is wishy-washy. It's simultaneously not specific enough to tell a tag from a branch with similar name and not general enough to point to a pull request `merge` ref.

[...]

PR #171, opened 2020-09-27
support git version starting with heads/ or tags/

dirk-thomas:
Implements part of #170.


dirk-thomas:
Without the second last commit (9365c6375e4ce8139b611c5fd2459aabbc5cb480) CI fails since one of the tests uses the `list.repos` file from `master` which is being modified as part of this PR. This commit temporarily modifies the URL to fetch the file from this branch to demonstrate that CI passes.

The last commit (263c529a49456e811b53af23e6531eb2c9f23de1) reverts the change again. While that will fail CI of this PR the merged PR should pass CI on master afterwards.

dirk-thomas:
The last commit (d4eb2b0f4dbd7c91e7de9d9ab545b7b072021a3d) ensures that CI passes even when `list.repos` is changed by using the remote URL pointing to the branch of the PR (instead of the default `master`).

mikepurvis:
Thanks for this change! I hadn't actually seen the original ticket, but having the option to be explicit about branches vs tags is really helpful to me.

PR #180, opened 2020-10-27
FIX: Test regressions on macOS

jenisys:
PROBLEMS:
* Subversion (svn) was not installed (but: test(s) should have been skipped)
* Mercurial (hg) was not installed (but: test(s) should have been skipped)
* git-version detection in macOS failed w/ existing implementation

CHANGES:
test/test_commands.py:
* Split-up test_validate() into two parts, one with git and one with svn and hg.
  REASON: Subversion and or Mercurial might not be installed.
  Therefore, skip this test if this is the case (where svn and hg are used).
* ADDED: test/program_finder.py to locate svn/hg programs (and check usability).

vcstool/client/git.py
* Adapted git-version detection becasue on macOS dit not work w/ installed Apple git.

test/test_flake8.py:
* error: unrecognized arguments: --import-order-style=google\n' => DISABLED
* Disabled pydocstyle => Caused import-errors (even if installed)

MORE:
* ADDED: py.requirements.txt to simplify to setup a workspace for the developer.
* ADDED: pytest.ini to be able to use pytest-html test-report (w/ detailled output)


jenisys:
flake8 tests suck (or suck on Travis): 
* How can that be that the flake8 tests pass on my computer and fail on Travis ?!?
* How can that be that the flake8 tests contains warnings/errors on things that I have not changed ?!?
* Where are the problematic outputs that I see on my computer (as long as things are not fixed) but not on Travis output ?!?
* Where is the flake8 config-file that states which checks/rules are used so that I am independent of flake8 version changes ?

dirk-thomas:
> FIX: Test regressions on macOS

Can you please elaborate what exactly has regressed on macOS? I don't see anything in the description which fixes a recently introduced regression.

> flake8 tests suck (or suck on Travis):

I would assume that your local setup / environment from the one on Travis (which is defined in the `.travis.yml` file).

---

Please don't use a single pull request to apply multiple unrelated patches. Instead please create separate pull requests for independent changes. That way they can be reviewed / commented on / merged individually.

I am happy to comment on individual pull requests with further feedback on the various parts of the patch.

jenisys:
The regressions are described in detail above (in the first commit):
* git-version retrieval failed on macOS, see the diffs in "git.py"
* pytests failed because modules could not be imported (did not find there symbols), etc.

The remaining commits are just trying to fix the CI-BUILDs.
When the CI-BUILD is OK again, I will provide a squashed commit for this pull-request.

REGARDING: The flake8 problems
My local setup is identical to the travis/github except that no version-constrains are specified (by you).
I am using a clean virtualenv to run my tests and everything is fine (after my fixes related to macOS and potentially other stuff).

dirk-thomas:
Since I expect the chances are slim that you will create separate pull requests for the unrelated changes as requested (based on the tone of the conversation) I went ahead and created a separate pull request to address the problem with the git version detection on macOS: see #181.

jenisys:
@dirk-thomas 
I close this PR due to the weird CI-BUILD problems that I encountered (flake8 tests run fine in my workspace).
Parts of this PR are provided as other PRs:

* PR #181 Fixes macOS git.get_version() problem (based on this PR; verified that this works on macOS by @jenisys)
* PR #182 Fixes the test regression when subversion/mercurial are not installed by skipping the test(s)

I could provide another PR for the remaining functionality (if this is desired):

* py.requirements.txt file(s) to simplify setup of a local workspace for developer(s)
* pytest.ini (with pytest-html) to generate an HTML-based test report (and optional an XML-based test report)
* tox.ini to simplify testing multiple Python versions in the developer workspace (to improve the feedback cycle on PRs)

ADDITIONAL COMMENTS:
* py.requirements.txt, ci-build scripts (travis.yml, github/*/ci.yml) should contain version constraints, like: `pytest >= 5.0`
* ci-build scripts (travis.yml, github/*/ci.yml) should use the upgrade mode for the required packages
  REASON: From my experience with Travis, it often provides outdated versions for some common packages, like pytest.
* ci-build scripts (for different CI providers) should use the same recipe. Currently, they differ slightly (and this should be cleaned up).
* A flake8 config-file should be used, to specify which rules are checked/important (decouple from any changed behaviour by another flake8 version)

dirk-thomas:
> PR #181 Fixes macOS git.get_version() problem (based on this PR; verified that this works on macOS by @jenisys)

Thanks for trying the patch and confirming that it works for you.

> py.requirements.txt file(s) to simplify setup of a local workspace for developer(s)

I appreciate the suggestion but I don't see a need for this. First the dependencies of this package are minimal and second the number of active developers is pretty low.

> pytest.ini (with pytest-html) to generate an HTML-based test report (and optional an XML-based test report)

Is there any particular information an HTML report would provide in addition? Without any additional value I don't think an html report is necessary - especially considering that this would add another test dependency. The current ASCII output seems sufficient to me.

> tox.ini to simplify testing multiple Python versions in the developer workspace (to improve the feedback cycle on PRs)

For a similar reason as above (this isn't a high bandwidth repo) I think relying on the GitHub workflow to test different versions is sufficient (at least for my workflow).

> ci-build scripts (travis.yml, github/*/ci.yml) should contain version constraints, like: pytest >= 5.0

Since CI always installs the latest version I don't see how specifying a lower boundary would be any different.

> ci-build scripts (travis.yml, github/*/ci.yml) should use the upgrade mode for the required packages

From my experience the CI Python environment never had any of these installed. But adding `--upgrade` to be on the safe side shouldn't harm. Done in #186.

> ci-build scripts (for different CI providers) should use the same recipe. Currently, they differ slightly (and this should be cleaned up).

GitHub workflows have only been recently added but have proven superior in terms of performance (and configuration). So I think the right move is to just remove Travis CI since currently all checks are redundant. Done in #185.

> A flake8 config-file should be used, to specify which rules are checked/important (decouple from any changed behaviour by another flake8 version)

The goal is that this package passes `flake8` in its default configuration. If that changes in the future then the package will follow with its style. Therefore CI runs on a cron-like schedule (beside on commits / pull requests) to catch failures from external changes like that.

jenisys:
RELATED TO: Weird test import errors
Weird pydocstyle problems with "test/test_flake8.py" were caused by a virtualenv bug/problem that may occur if "pydocstyle" (and flake8, ...) are not installed in the underlying python implementation (weird problem that core modules are used and only import from core installation instead of virtual-environment).

PR #182, opened 2020-10-28
allow some tests (except on CI) to be skipped when svn/hg are unavailable

jenisys:
Split-up test_validate() into two parts:
* test_validate: Part requires only git (subtest 1 and subtest 3)
* test_validate2: Part requires subversion (svn) and mercurial (hg) (former middle part = subtest 2)
* test_validate2 is skipped if either svn or hg are not installed 
  (skip logic was manually tested: with svn and hg, with svn, with hg, without svn and without hg).
* support: FAKE-PROGRAM detection that exist (svn)
  but only states that the program is missing if it is run
  (usecase: macOS for svn that was a former part of Xcode).

HINT: PR #181 works for me (in my workspace).

### DIFF
--- a/test/test_commands.py
+++ b/test/test_commands.py
@@ -16,14 +16,16 @@ REPOS2_FILE = os.path.join(os.path.dirname(__file__), 'list2.repos')
 TEST_WORKSPACE = os.path.join(
     os.path.dirname(os.path.dirname(__file__)), 'test_workspace')
 
-CI = os.environ.get("CI", "false") == "true"  # Travis, Github, ... set: CI=true
-svn = which('svn') or False
-hg = which('hg') or False
-if svn and os.system('{svn} --version'.format(svn=svn)):
-    # -- CHECK FOR FAKE-PROGRAM (usecase: macOS, ...)
-    # FAKE-PROGRAM only states that program is not installed (and what to do).
-    # FAKE-PROGRAM has normally EXIT_FAILED exitcode (non-zero).
-    svn = False
+CI = os.environ.get('CI') == 'true'  # Travis CI / Github actions set: CI=true
+svn = which('svn')
+hg = which('hg')
+if svn:
+    # check if the svn executable is usable (on macOS)
+    # and not only exists to state that the program is not installed
+    try:
+        subprocess.check_call([svn, '--version'])
+    except subprocess.CalledProcessError:
+        svn = False
 
 
 class TestCommands(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> ... regression  ...

As mentioned on the previous ticket I don't think this is a regression. The tests have relied on `svn` and `hg` forever. Or is there a specific commit which broke this which I might not be aware of?

I am not sure this is the best way to handle the case if either `svn` or `hg` are not present. Basically if either of the them are missing the test suite results are incomplete. That also implies if for whatever reason they are unavailable in CI the result might still be reported as "all ok" even though parts of the code aren't being tested. Therefore I think tests *should* fail if the necessary dependencies aren't present. What is the current error message / test failure you are getting? Maybe they can be improved so it is easy to see why the tests are failing and if the developers doesn't have them installed they can choose to ignore the test failure.

jenisys:
Mmh, there is a fundamental difference between:
* CASE_1: tests that fail because the assertions/verifications are not met (meaning you xxx-UNDER-TEST is broken)
* CASE_2: tests that fail because their assumptions (preconditions) are not met

Regarding CASE_2 (when assumptions are not met):
* Do you gain any insights about your PACKAGE-UNDER-TEST if you run the tests (when the assumptions are not met) ? 
  ANSWER: NO, they will always fail.
* Does that mean that your PACKAGE-UNDER-TEST is broken ? 
  ANSWER: NO, but test coverage will indicate that you have not covered the complete functionality.
* If you run the test and it fails how easy it is to determine, if CASE_1 or CASE_2 caused the failure ?
* Have you somewhere stated/described (README or other docs) what your test assumptions are ? No (AFAIK)
* If you user runs your test suite and it fails (CASE_2), will he/she come to you and say something is broken (and waste your time) ?

Therefore, CASE_2 is normally skipped because:
* You explicitly state your assumptions in the test (it is readable)
* It makes your test suite more robust / resilient against user error(s)
* It helps you in complex test beds where each test supports only a part of all needed features (auto-detects/-configures what works and what not). 
 
IN ADDITION:
Yes, your test suite passes, but with SKIPPED tests mentioned in the test summary (just look at the summary).
In your case, you will always have all the supported version control systems installed.
For other developers that may not be the case.

Also, the change is rather minimal, which indicates how important the "other version control systems" currently are you and the users of your package.

dirk-thomas:
I agree with the majority of your statements. Marking those tests as skipped is the right semantic.

> Yes, your test suite passes, but with SKIPPED tests mentioned in the test summary (just look at the summary).

This is the point were I have to disagree though. If CI passes (with or without skipped tests) then the status will be green signaling that everything is ok which is potentially not the case. So if the test is only marked "skipped" and doesn't fail then CI **must** consider such a skipped test as a failure to avoid letting the lack of testing go unnoticed.

And when having to choose I certainly prefer safe defensive CI results over the inconvenience of a developer which doesn't have all vcs types installed to manually ignore tests related to that.

jenisys:
But the problem that CI should not skip tests is easy.
You just need to detect that your run as part of a CI build (via an environment variable, etc.) and make that part of the boolean-skip-expression that decides if the test should run or not.

dirk-thomas:
> You just need to detect that your run as part of a CI build (via an environment variable, etc.) and make that part of the boolean-skip-expression that decides if the test should run or not.

That sounds good to me. `CI` you should be set on Travis CI as well as GitHub actions.

dirk-thomas:
(Rebased after #190 was merged to run GitHub actions.)

dirk-thomas:
Thanks for the patch.

PR #183, opened 2020-10-28
Add fish autocompletion

bergercookie:
Current PR adds the following:

A fish autocompletion file and the `--commands-descriptions` flag for printing the vcs subcommands in a tab-separated list along with their descriptions 
~* A small change in the README suggesting the use of `pip -e` instead of `sudo python setup.py` for installing in dev mode~
~* Minor reformatting in the flow control of the CLI parsing~

Here's how the autocompletion looks in fish now:

1603867624-screenshot
1603867610-screenshot

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -191,6 +191,9 @@ For *zsh* append the following line to the ``~/.zshrc`` file::
 
   source /usr/share/vcstool-completion/vcs.zsh
 
+For *fish* append the following line to the ``~/.config/fishconfig.fish`` file::
+
+  source /usr/share/vcstool-completion/vcs.fish
 
 How to contribute?
 ==================
--- a/vcstool/commands/help.py
+++ b/vcstool/commands/help.py
@@ -36,6 +36,11 @@ def main(args=None, stdout=None, stderr=None):
         print(' '.join([cmd.command for cmd in vcstool_commands]))
         return 0
 
+    if ns.commands_descriptions:
+        print('\n'.join(['{}\t{}'.format(cmd.command, cmd.help)
+                         for cmd in vcstool_commands]))
+        return 0
+
     # output detailed command list
     parser = get_parser_with_command_only()
     parser.print_help()
@@ -57,6 +62,9 @@ def get_parser(add_help=True):
     group.add_argument(
         '--commands', action='store_true', default=False,
         help='Output the available commands for auto-completion')
+    group.add_argument(
+        '--commands-descriptions', action='store_true', default=False,
+        help='Output the available commands along with their descriptions')
     from vcstool import __version__
     group.add_argument(
         '--version', action='version', version='%(prog)s ' + __version__,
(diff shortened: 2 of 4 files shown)


README.rst
```diff
@@ -210,5 +213,7 @@ How to try the latest changes?
 Sourcing the ``setup.sh`` file prepends the ``src`` folder to the ``PYTHONPATH`` and the ``scripts`` folder to the ``PATH``.
 Then vcstool can be used with the commands ``vcs-COMMAND`` (note the hyphen between ``vcs`` and ``command`` instead of a space).
 
-Alternatively the *develop* command from Python setuptools can be used:
-  sudo python setup.py develop
+Alternatively the ``-e/--editable`` flag of ``pip`` can be used::
+
+  # from the top level of this repo
+  pip3 install --user -e .
```
dirk-thomas, line 219: Please create a separate pull request for this unrelated change. Please also describe why this should be changed.

vcstool/commands/help.py
```diff
@@ -30,15 +30,16 @@ def main(args=None, stdout=None, stderr=None):
...
+        print('\n'.join(['{}\t{}'.format(cmd.command, cmd.help)
+                         for cmd in vcstool_commands]))
+    else:
+        # output detailed command list
+        parser = get_parser_with_command_only()
+        parser.print_help()
 
-    # output detailed command list
-    parser = get_parser_with_command_only()
-    parser.print_help()
```
dirk-thomas, line 41: Please keep the current structure of the code (`if ...: ...; return 0`) which should minimize the changes.

bergercookie: I thought of refactoring this since I believe the `if .. elif .. else` conditional is easier to read; I'll revert it

README.rst
bergercookie: OK, See https://github.com/dirk-thomas/vcstool/pull/184/files

vcstool/commands/help.py
bergercookie: done

dirk-thomas: > I thought of refactoring this since I believe the if .. elif .. else conditional is easier to read;

Certainly subjective :wink: Imo it is easier to see that when an `if args.xxx:` block contains a return I don't have to look further down if it does anything else.

dirk-thomas: It looks like that the updated patch doesn't handle `args.commands_descriptions` anymore?

bergercookie: Ouch, rushed it..

dirk-thomas: No problem. Thanks for the quick updates!

dirk-thomas:
Thanks for the improvement.

dirk-thomas:
@bergercookie Just FYI #191.

bergercookie:
Ah, yeah, my bad;
Are there any tests for vcstool, e.g., in the ROS buildfarm?
If so we could add a test for the `setup.py install` scenario

dirk-thomas:
This repo does contain tests but they weren't triggered for this PR. We recently disabled Travis CI in favor of GitHub actions but until #190 the latter were not triggered for PRs coming from forked repos.

PR #184, opened 2020-10-28
Don't use sudo with pip

bergercookie:
Using sudo directly with `pip` is considered a bad practice in the python community, since  it's rarely actually needed, it may mess up file permissions and it can actually break `pip` if used incorrectly:

See following links for more details:

- https://github.com/pypa/pip/issues/5495
- https://askubuntu.com/a/1149202/305258

Also using sudo in this case, is unnecessary since we're operating in development mode (i.e., not copying the file contents systemwide)


dirk-thomas:
Thanks for the improvement.

PR #203, opened 2021-03-20
Detect special POSIX shells on Windows and output POSIX paths

### LINKED ISSUE #202: Add --repos-posix option to force POSIX paths
This is a feature request. I'd be happy to open a PR after getting an OK.

In some cases, on Windows, users might have to work with POSIX paths (i.e. provide POSIX paths and expect POSIX paths). However, providing a POSIX path (as a base path) on Windows results in paths with a POSIX/Windows mix. For example:

```
$ vcs diff src/ -s --repos
src/ament\ament_cmake (git)
...
```

This is annoying to have to deal with if we intend to pipe the resulting paths into something else. For a concrete example, see https://github.com/ros-tooling/action-ros-ci/issues/578.

Therefore I am proposing to add a `--repos-posix` option (as a common argument) that would be just like `--repos` but converting repo paths to POSIX paths.

Something like:

```diff
diff --git a/vcstool/executor.py b/vcstool/executor.py
index 9b0b301..ea36a1f 100644
--- a/vcstool/executor.py
+++ b/vcstool/executor.py
@@ -1,4 +1,5 @@
 import logging
+import pathlib
 import os
 from queue import Empty, Queue
 import sys
@@ -9,12 +10,13 @@ logger = logging.getLogger(__name__)
 logging.basicConfig()


-def output_repositories(clients):
+def output_repositories(clients, force_posix=False):
     from vcstool.streams import stdout
     ordered_clients = {client.path: client for client in clients}
     for k in sorted(ordered_clients.keys()):
         client = ordered_clients[k]
-        print('%s (%s)' % (k, client.__class__.type), file=stdout)
+        path = pathlib.Path(k).as_posix() if force_posix else k
+        print('%s (%s)' % (path, client.__class__.type), file=stdout)


 def generate_jobs(clients, command):
```

A more "global" alternative could be a `--posix` option that would affect paths printed by both `output_repositories()` (for `--repos`) and `output_result()`/`output_results()`.

christophebedard:
Closes #202

The real problem is that some special shells on Windows may not support backslashes at all.

* PowerShell supports mixed slashes
* CMD mostly doesn't
* bash (installed along with Git for Windows) only supports forward slashes

I've tried to find a simple method to detect special shells like this, since we preferably only want to change the current behaviour for those shells, although it's okay to change the behaviour for shells that support mixed styles. I looked into `MSYSTEM` but ultimately found that checking if `$_` exists and contains a forward slash seems to work fine.

> **Special Parameters**
> **_**
> At shell startup, set to the absolute pathname used to invoke the shell or shell script being executed as passed in the environment or argument list. Subsequently, expands to the last argument to the previous command, after expansion. Also set to the full pathname used to invoke each command executed and placed in the environment exported to that command.  [...]
> https://linux.die.net/man/1/bash

Note that this PR only affects the paths that are printed (1) as part of the results and (2) when using the `--repos` option. I didn't want to change anything else (e.g. applying the path fix in the crawling step), because that seems to be working fine the way it is.

Finally, as a follow-up to the concrete downstream issue described in https://github.com/ros-tooling/action-ros-ci/issues/578, I did try the fix from this PR and it works. The bash shell is detected on Windows and `vcstool` prints POSIX paths. I also tested manually, and it doesn't change the current behaviour for PowerShell or CMD.

### DIFF
--- a/vcstool/executor.py
+++ b/vcstool/executor.py
@@ -9,12 +9,21 @@ logger = logging.getLogger(__name__)
 logging.basicConfig()
 
 
+# Detect special Windows shells that do not support mixes of
+# backslashes & forward slashes; for those shells, we want to
+# output POSIX paths, i.e. forward slashes only
+windows_force_posix = \
+    sys.platform == 'win32' and '/' in os.environ.get('_', '')
+def fix_output_path(path):
+    return path.replace('\\', '/') if windows_force_posix else path
+
+
 def output_repositories(clients):
     from vcstool.streams import stdout
     ordered_clients = {client.path: client for client in clients}
     for k in sorted(ordered_clients.keys()):
         client = ordered_clients[k]
-        print('%s (%s)' % (k, client.__class__.type), file=stdout)
+        print('%s (%s)' % (fix_output_path(k), client.__class__.type), file=stdout)
 
 
 def generate_jobs(clients, command):
@@ -70,6 +79,8 @@ def execute_jobs(
     from vcstool.streams import stdout
     if debug_jobs:
         logger.setLevel(logging.DEBUG)
+
+    logger.debug('force POSIX paths on Windows: %s' % windows_force_posix)
 
     results = []
 
@@ -223,7 +234,7 @@ def output_result(result, hide_empty=False):
         client = result['client']
         print(
             ansi('bluef') + '=== ' +
-            ansi('boldon') + client.path + ansi('boldoff') +
+            ansi('boldon') + fix_output_path(client.path) + ansi('boldoff') +
             ' (' + client.__class__.type + ') ===' + ansi('reset'),
             file=stdout)
     if output:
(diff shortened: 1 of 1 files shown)


christophebedard:
As for tests, I have written one, but I'd also need to make sure that bash is installed in the testing environment and all. Should I proceed?

Also, it seems that tests are broken for Ubuntu Focal/possibly a newer Git version, and I don't think the tests work on Windows.

dirk-thomas:
Thank you for the patch.

I rebased your commit on top of the latest state of the target branch to fix unrelated CI failures. Additionally I addressed the linter failures in a separate commit ([670b017](https://github.com/dirk-thomas/vcstool/pull/203/commits/670b0172db4854eb5222a881c9d27aecad264c16)) as well as made the debug message only conditional when the logic is enabled ([462007e](https://github.com/dirk-thomas/vcstool/pull/203/commits/462007eb28bb5a0229b4e71b31d7521191bdc97e)).

PR #206, opened 2021-04-07
Fix #205: `--recursive` is not recursive 

### LINKED ISSUE #205: `--recursive` is not recursive
https://github.com/dirk-thomas/vcstool/blob/ea4c777989fbdc71d0163beb0ec6ea69f524a70d/vcstool/clients/git.py#L441-L443

The current implementation is to execute `git submodule update --init` when `vcs import --recursive` is executed, but this does not perform recursive processing.
To do recursive processing, it is needed to add `--recursive`, as in `git submodule update --init --recursive`.

eduidl:
fix: https://github.com/dirk-thomas/vcstool/issues/205

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -440,7 +440,8 @@ class GitClient(VcsClientBase):
 
         if command.recursive:
             cmd_submodule = [
-                GitClient._executable, 'submodule', 'update', '--init']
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch.

I just rebased your commit to the latest state of the target branch to address the unrelated CI failures.

PR #211, opened 2021-05-09
Make CI jobs use matrix.os value and fix tests for macOS & Windows

christophebedard:
There was a strategy matrix, but all jobs were running on `ubuntu-latest`.

Tests run fine on macOS if I set `advice.detachedHead` to `true` (I think it's off by default on newer git versions, and macOS has a more recent version).

For Windows, it's a bit more complicated because of the different path separator. It required some more manual adjustments to the test code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest]
@@ -33,4 +33,4 @@ jobs:
       run: |
         pip install --upgrade coverage flake8 flake8-docstrings flake8-import-order pytest
         git config --global --add init.defaultBranch master
-        PYTHONPATH=`pwd` pytest -s -v test
+        ${{ matrix.os == 'windows-latest' && 'set PYTHONPATH=%cd% &&' || 'PYTHONPATH=`pwd`' }} pytest -s -v test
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Nice catch. Thank you for this patch.

> I tried to get tests to pass on Windows (christophebedard@c6880c5), but it's a bit of a nightmare because of the path separator. I've simply disabled it.

The Windows specific patch seems reasonable. Did the tests pass with it or where there other problems?

If they passed, I would suggest to add that snippet to this PR. If you first replace the cwd and the change the path separators you can avoid having to change the path separators of the cwd. Also wrapping those new lines in a if-windows-condition would be probably be good.

christophebedard:
> Did the tests pass with it or where there other problems?

No they didn't pass with that patch unfortunately. I don't remember the exact details, but I think find-and-replace just wasn't enough.

> If you first replace the cwd and the change the path separators you can avoid having to change the path separators of the cwd. Also wrapping those new lines in a if-windows-condition would be probably be good.

I'll try that.

christophebedard:
> > If you first replace the cwd and the change the path separators you can avoid having to change the path separators of the cwd. Also wrapping those new lines in a if-windows-condition would be probably be good.
> 
> I'll try that.

Actually I'm not sure I understand what you mean. The two `output.replace` calls are for two different cases.

In any case, I changed it a bit and fixed the remaining issues. It's not pretty.. but it works™️ 😆

I'm going to update the PR title & description.

dirk-thomas:
Thank you for working on this significant fix for the test coverage.

I just rebased the patch and slightly altered it to minimize the diff and keep comparing strings (instead of bytes).

Issue #195, opened 2020-11-16
Add `remove` command


bergercookie:
I see `vcstool` as a good alternative to git modules in case one needs to have multiple available configurations YAML configurations in the same git repository and branch. For example a project may have two YAML files, one for the dependencies that need to be installed e.g. for ROS melodic, and  a second slightly different set of dependencies for noetic. In such cases, if one wants to alternate between the two, it would be convenient to offer a subcommand like `vcs remove`, which would remove the cloned submodules of a previous run, so that the next `vcs import ...` is done with a cleaner slate.

Would such a functionality make sense?
Is there something like this already implemented that I'm missing?

dirk-thomas:
I am not sure I understand the use case. If you have to different sets of repositories why don't you clone them into two different locations in the first place? And if you want to use the same location why can't you simply delete the existing working copies before import the other set?

Also it is a unclear what an `import` verb should do - maybe you can describe that with a list of commands you would intend to invoke and their desired behavior.

bergercookie:
> And if you want to use the same location why can't you simply delete the existing working copies before import the other set?

You could remove them by hand, yes. But if there are numerous repositories handled by `vcs` then the user would have to form a lengthy command of the form `rm -rf src/... ... `. Also if some of the items / repositories under `src` were not managed by `vcs` in the first place, the user would need to take extra care not to remove them.
Definitely all these are doable, but an `rm-all` verb would be convenient  to remove all the vcs-managed repositories automatically and avoid the above issues

> Also it is a unclear what an import verb should do - maybe you can describe that with a list of commands you would intend to invoke and their desired behavior.

I'd add a single verb `rm-all` which would remove all vcs-managed repositories under the given path. The user would call it given the path as well as the `YAML` file from which to read the list of repositories to remove `vcs rm-all src < <path/to/repos.yaml>` 

I don't plan to modify the existing `import` verb

dirk-thomas:
> I'd add a single verb `rm-all` which would remove all vcs-managed repositories under the given path. The user would call it given the path as well as the `YAML` file from which to read the list of repositories to remove `vcs rm-all src < <path/to/repos.yaml>`

Sounds good to me. It would be good to also allow an option `--input` to match the `import` arguments.

I would suggest to ask for confirmation after listing the to-be-deleted directories by default. An option like `--yes` could then be used to skip the interactive confirmation.

dirk-thomas:
[...]

PR #216, opened 2021-05-26
Add an `rm-all` option to `vcs`

bergercookie:
As discussed earlier in https://github.com/dirk-thomas/vcstool/issues/195 it would be useful to have an `rm-all` option whose purpose is to remove the already cloned repositories, as specified by the given YAML file. This is so that the user can potentially clone a new set of dependencies using a new YAML file and not have to manually clean the target directory.

`rm-all` has a similar interface to `import` in that it accepts both `--input ...` as well as input redirection.
It also requires either the `-f/--force` option or the `-n/--dry-run` option to actually execute. This is the same behavior as tools such as `git clean` and its purpose is to avoid a user mistake.

Here are some example usages of the command:

Typical usage:
```sh
$ vcs rm-all -f src/  < repos.yaml
Paths to delete:

- src/follow_waypoints
- src/kobuki
```

```sh
$ vcs rm-all -n src/  < repos.yaml
Paths to delete:

- src/follow_waypoints
- src/kobuki

[Dry Run]
```

Not allowed:
```sh
$ vcs rm-all  src/  < repos.yaml
usage: vcs rm-all [-h] [--input FILE_OR_URL] (-n | -f) [path]
vcs rm-all: error: one of the arguments -n/--dry-run -f/--force is required
```


bergercookie:
@dirk-thomas , friendly ping. :)

bergercookie:
@dirk-thomas , friendly ping

LukeAI:
I'd also like to get this feature, but it looks like there's no active development happening here anymore, could be worth considering PRing to https://github.com/MaxandreOgeret/vcstool2/

bergercookie:
Hi @LukeAI,

I don't currently plan putting more effort in this PR. Do feel free however to port this to the new vcstool2 repo. ;)

LukeAI:
> Hi @LukeAI,
> 
> I don't currently plan putting more effort in this PR. Do feel free however to port this to the new vcstool2 repo. ;)

fair enough! thanks, sounds like a good weekend project.

Issue #218, opened 2021-06-15
[feature-request] Support sparse-checkout


aaronplusone:
git 2.25+ allows tracking of partial repository directories through [sparse-checkout](https://git-scm.com/docs/git-sparse-checkout). This is useful when a project needs only a limited set of packages in a monolithic repository. E.g. `swri_yaml_util` in [swri-robotics/marti_common](https://github.com/swri-robotics/marti_common).

I've been working on a branch to adapt vcs to do sparse-checkouts for our CI/CD pipeline. If you approve of the approach and are comfortable accepting git specific functionality, I can make a PR.

#### YAML
* New "subpaths" key specifies sparse-checkout patterns for repository. Key is the same level as `url/uri`.

Example from above, with a second package specified.

```yaml
# VCS Format
repositories:
  marti:
    type: git
    url: https://github.com/swri-robotics/marti_common.git
    version: master
    subpaths:
      - swri_yaml_util
      - swri_console_util
```

#### import_.py
* Add subpaths member to ImportCommand
* Parsing logic for subpaths in `get_repos_in_vcstool_format` and `get_repos_in_rosinstall_format`
* Include subpaths in `generate_job`

#### Git Client
* Add check for git version > 2.25.0 when "subpaths" key is found in import/export
* Import
    + Match sparse-checkout patterns when determining if existing repo matches yaml specification
    + If subpaths in command
        - Clone with `--no-checkout`
        - Initialize sparse repo with `--cone`
        - Set sparse-checkout patterns to space separated list of subpaths
        - Verify that subpaths exist in final repo, warn if they do not (to protect against typos in subpath specification)
* Export
  + Add subpaths (if any exist) to `export_data` key

130s:
+1

I discussed offline with @aaronplusone  and suggested:
- Backup compatibility: i.e. the tool should not fail when 1) the option to trigger the new feature is not passed, 2) the underlining functionality is not available in non-git `vcs` tool.
   - For non-`git`, handle unimplemented feature and behave the same way that the tool works without the option.

aaronplusone:
> +1
> 
> I discussed offline with @aaronplusone and suggested:
> 
>     * Backup compatibility: i.e. the tool should not fail when 1) the option to trigger the new feature is not passed, 2) the underlining functionality is not available in non-git `vcs` tool.
>       
>       * For non-`git`, handle unimplemented feature and behave the same way that the tool works without the option.

[...]

PR #219, opened 2021-06-15
Feature sparse checkouts

aaronplusone:
Addresses https://github.com/dirk-thomas/vcstool/issues/218

Adds accommodation for `git sparse-checkout`.

Tested on ubuntu:20.04/git=2.25.1 and in docker ubuntu:18.04/git=2.32.0.


christophebedard:
Adding some tests would be great I think.

vcstool/clients/git.py
```diff
@@ -497,6 +561,23 @@ def _get_remote_urls(self):
             'returncode': 0 if remote_urls else 1
         }
 
+    def _get_sparse_patterns(self):
+        cmd = [GitClient._executable, 'sparse-checkout', 'list']
+        result = self._run_command(cmd)
+        if result['returncode']:
+            result['output'] = 'Could not list sparse patterns: ' \
```
130s, line 568: Not sure if this user-facing message, but if it is, then don't we want to give more hints for why?

Looking at what `git sparse-checkout list` returns, I guess this `if` block hits when the repo is not set for `sparse-checkout`?

```
$ git sparse-checkout list
warning: this worktree is not sparse (sparse-checkout file may not exist)

$ git sparse-checkout set pkg_c
$ ll
total 16K
drwxr-xr-x 4 n130s n130s 4.0K Jun 16 13:57 ./
drwxr-xr-x 8 n130s n130s 4.0K Jun 16 13:57 .git/
drwxrwxr-x 4 n130s n130s 4.0K Jun 16 13:57 pkg_c/
drwxrwxr-x 3 n130s n130s 4.0K Jun 16 13:57 ../
$ git sparse-checkout list
pkg_c
```

README.rst
```diff
@@ -50,7 +50,8 @@ Exporting and importing sets of repositories
 Vcstool can export and import all the information required to reproduce the versions of a set of repositories.
 Vcstool uses a simple `YAML <http://www.yaml.org/>`_ format to encode this information.
 This format includes a root key ``repositories`` under which each local repository is described by a dictionary keyed by its relative path.
-Each of these dictionaries contains keys ``type``, ``url``, and ``version``.
+Each of these dictionaries contains keys ``type``, ``url``, and ``version``. The optional ``subpaths`` key may be used
+to specify `sparse-checkout <https://git-scm.com/docs/git-sparse-checkout>`_ directories as a list.
```
130s, line 54: Maybe clarify this is (currently) available to `git` only? Assuming majority usecases are with `git`, I don't know how important though.

README.rst
```diff
@@ -50,7 +50,8 @@ Exporting and importing sets of repositories
 Vcstool can export and import all the information required to reproduce the versions of a set of repositories.
 Vcstool uses a simple `YAML <http://www.yaml.org/>`_ format to encode this information.
 This format includes a root key ``repositories`` under which each local repository is described by a dictionary keyed by its relative path.
-Each of these dictionaries contains keys ``type``, ``url``, and ``version``.
+Each of these dictionaries contains keys ``type``, ``url``, and ``version``. The optional ``subpaths`` key may be used
```
130s, line 53: No-blocking, just saying; Since underlining functionality (in `git`) is `sparse-checkout`, naming it differently may not be intuitive. However, `subpath` makes more sense in the context of attribute name `.repos` file (and in this context, `sparse-checkout` wouldn't sound reasonable).

vcstool/clients/git.py
```diff
@@ -439,6 +468,27 @@ def import_(self, command):
 
                 checkout_version = command.version
 
+        if command.subpaths:
```
130s, line 471:
```suggestion
        if command.subpaths:
            # Setting the paths sparse-checkout deals with.
```
Particularly because I see `if command.subpaths` block appears multiple times in this method, I think briefing the purpose per each chunk is helpful.

130s (changes requested):
- Change suggested: Unit tests will be helpful, particularly when [git-scm.com](https://git-scm.com/docs/git-sparse-checkout) mentions it's experimental.
   > THIS COMMAND IS EXPERIMENTAL. ITS BEHAVIOR, AND THE BEHAVIOR OF OTHER COMMANDS IN THE PRESENCE OF SPARSE-CHECKOUTS, WILL LIKELY CHANGE IN THE FUTURE.
- Just a question: For local repo, does the `vcstool` with this change always synch sparse setting status? What if the sparse setting is different b/w local and remote repos? I couldn't figure that out from the code.
- Not-blocking. I'd be a bit worried that the version requirement for git 2.25 scattered in multiple places would make maintenance harder.

vcstool/clients/git.py
```diff
@@ -197,13 +205,20 @@ def export(self, command):
                 url = result_url['output']
                 cmds.append(result_url['cmd'])
 
+                if self._check_sparse():
```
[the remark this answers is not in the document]
aaronplusone: How about moving it above the exact if/else?

README.rst
aaronplusone: The documentation refers to them as "patterns", which I use elsewhere in the code.

> Write a set of patterns to the sparse-checkout file, as given as a list of arguments following the set subcommand. Update the working directory to match the new patterns.

I could be more explicit and call them `sparse-checkout-patterns`. This could lead to more complexity since in the git call I used `--cone` to restrict patterns to directories only (see https://git-scm.com/docs/git-sparse-checkout#_cone_pattern_set), and I'd either need a way to indicate that non-cone patterns aren't supported or I'd need to add support for non-cone clones.

aaronplusone: Another note, though, is that this type of action could be supported in other vcs. Calling it after the git specific command could cause problems in the future.

vcstool/clients/git.py
aaronplusone: The pattern I saw elsewhere is to list the previous commands' output following the diagnostic. Looks like I omitted the key, but it's there now.

This also brings up an interesting point on having a sparse-checkout without any patterns. The command will return with 0. I suppose this could be a valid use case if you only want the top level of a repository, so I'll rewrite to ensure we don't fail with an empty pattern list.

130s:
> Just a question: For local repo, does the `vcstool` with this change always synch sparse setting status? What if the sparse setting is different b/w local and remote repos? I couldn't figure that out from the code.

Responding myself:
- I think `sparse-checkout` only modifies the status of local repo? So remote repos always retain full, non-sparse state?
- That said, what if a user 1) clones a repo by this suggested feature, 2) modifies the `subpath` entries locally using `git sparse-checkout`, then 3) run again vcstool's command enabling sparse checkout? I guess that might be out of scope of this PR.

130s:
Not sure why CI hasn't run.

For almost all recent PRs CI seems to have run https://github.com/dirk-thomas/vcstool/pulls?q=is%3Apr+

christophebedard:
> Not sure why CI hasn't run.
> 
> For almost all recent PRs CI seems to have run [https://github.com/dirk-thomas/vcstool/pulls?q=is%3Apr+](https://github.com/dirk-thomas/vcstool/pulls?q=is%3Apr+sort%3Aupdated-desc)

GitHub added this very recently (in the last week?):

> First-time contributors need a maintainer to approve running workflows. [Learn more.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

In the meantime, you can probably run CI in a PR on a fork though

aaronplusone:
> > Just a question: For local repo, does the `vcstool` with this change always synch sparse setting status? What if the sparse setting is different b/w local and remote repos? I couldn't figure that out from the code.
> 
> Responding myself:
> 
>  I think `sparse-checkout` only modifies the status of local repo? So remote repos always retain full, non-sparse state?

Correct

>  That said, what if a user 1) clones a repo by this suggested feature, 2) modifies the `subpath` entries locally using `git sparse-checkout`, then 3) run again vcstool's command enabling sparse checkout? I guess that might be out of scope of this PR.

When checking if the repository exists, I also check the repo config for `core.sparsecheckout` to identify whether the local repo is sparse. If it is, I compare patterns and use the same logic as for differing repos. I.e., throw an error with no other flags, skip if `--skip-existing` flag is passed, and overwrite if `--force` flag is passed.

aaronplusone:
> > Not sure why CI hasn't run.
> > For almost all recent PRs CI seems to have run [https://github.com/dirk-thomas/vcstool/pulls?q=is%3Apr+](https://github.com/dirk-thomas/vcstool/pulls?q=is%3Apr+sort%3Aupdated-desc)
> 
> GitHub added this very recently (in the last week?):
> 
> > First-time contributors need a maintainer to approve running workflows. [Learn more.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)
> 
> In the meantime, you can probably run CI in a PR on a fork though

Ran against my fork, but the svn and hg tests are failing due to the icedtea8 link. http://icedtea.classpath.org/hg/icedtea8 results in a 404. Happy to include a different path for those tests if someone can direct me to it, but it's outside the scope of the PR.

aaronplusone:
As this is going on 2 years with no progress, I'm going to close this to clean up.

tfoote:
A few pieces of feedback for this that I posted in the [Discourse thread](https://discourse.ros.org/t/vcstool-has-been-stale/33661/8?u=tfoote) and reposting here for better discoverability

A few challenges with this is that only supported on some platforms (git >= 2.25) so is only supported on some of our target platforms. It also is only implementing a command for a specific vcs type of 4. It also is changing a moderately stable file format. And the workaround is to just checkout the whole tree. Or to specifically cherry pick the content out and cache it elsewhere. And lastly a common workaround for released packages is to simply clone from the release repo's branch: https://github.com/ros2-gbp/marti_common-release/tree/release/rolling/swri_console_util or tagged export https://github.com/ros2-gbp/marti_common-release/tree/release/rolling/swri_console_util/3.6.1-1 

The sparse checkout doesn't completely fit with the general git model of work. Consequently the [git spare-checkout](https://git-scm.com/docs/git-sparse-checkout) is explicitly an Experimental feature with the warning in the docs
"THIS COMMAND IS EXPERIMENTAL. ITS BEHAVIOR, AND THE BEHAVIOR OF OTHER COMMANDS IN THE PRESENCE OF SPARSE-CHECKOUTS, WILL LIKELY CHANGE IN THE FUTURE." Which makes it something that we don't necessarily want to silently turn on/recommend for users.

aaronplusone:
@tfoote, thanks for the feedback and for taking the time to cross-post. 

Your challenge points are certainly valid, and I wish I'd had that discussion when I originally worked on the feature. It only targeted my use case, used experimental features that were still relatively new, and had reasonable alternative approaches with minor trade-offs. If I recall correctly, this PR also came right around the time @dirk-thomas was transitioning to Nvidia, so there was some bad timing, too.

On the note about supported platforms (git >= 2.25), that was true at the time I wrote this, but is it still? Melodic went EoL before I closed the PR and Noetic targets Ubuntu 20.04 which comes with git 2.25.1.

Also, with respect to this feature only targeting git, that same concern came up with another of my ROS [PRs](https://github.com/ros/rosdistro/pull/28676#pullrequestreview-619092764) from that time. ROS has a vast target base, and covering the edge cases is difficult, time consuming, and an ever moving target. I think that's probably the reason why workarounds are so common in the ROS ecosystem.

I would like to think that encouraging smaller contributions helps maintain and improve official tools and reduces splintering (e.g., `catkin_make` vs `catkin tools` vs `colcon`), but it can also cause issues with ensuring code quality and user experience. Not sure how I'd weigh the balance if I was maintaining the official ROS tool instead of adapting it for a midsize company.

berndpfrommer:
This feature would have come in handy for me today. Sorry it never saw the light of day. I have a large repo that doesn't compile in its entirety on some distros, but I only needed the "msgs" package from the repo. Oh well. The workaround was to not compile on the distros where the compilation fails.

aaronplusone:
@berndpfrommer, you're welcome to use the Plus One Robotics [fork](https://github.com/plusone-robotics/vcstool) if @tfoote's workarounds won't work for you. We've been using it internally for the past year and a half.

You're also welcome to take the patch directly from the [source branch](https://github.com/aaronplusone/vcstool/tree/feature-sparse-checkouts) for your own purposes.

PR #239, opened 2022-04-22
Fix: UnboundLocalError: local variable 'version_type' referenced before assignment

furushchev:
`vcs validate` fails if version is referred to git hash due to unbounded variable error like below.

```
=== <package> (git) ===
Invocation of command 'validate' on client 'git' failed: UnboundLocalError: local variable 'cmd' referenced before assignment (/usr/lib/python3/dist-packages/vcstool/clients/git.py:736)
```

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -725,6 +725,8 @@ class GitClient(VcsClientBase):
             else:
                 for _hash in hashes:
                     if _hash.startswith(command.version):
+                        version_type = 'hash'
+                        version_name = command.version
                         break
                 else:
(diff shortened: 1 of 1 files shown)


EWouters:
I experienced this error as well, but this doesn't correctly fix the issue. In my case with the patch (3c0d37078f2ad35c70c5d0a2e125a08f7f2f04d9) it will return a hash, while my repo does have a tag checked out.

It turns out my repo has multiple tags, A and B, that point to the same commit. In the `repos.yaml` file tag A is requested but the repo seems to report tag B is checked out.

PR #256, opened 2023-03-26
Provide clearer error message when repository entry is empty

### LINKED ISSUE #255: bad non-descriptive error message 'NoneType' object is not subscriptable
This error message should be improved so it doesn't scare people, perhaps by describing exactly what data is set to None and maybe the line number of the input file.

Here is an example of a malformatted .repo file:
```yaml
# example.repo
repositories:
  bad:
  ament_cmake:
    type: git
    url: https://github.com/ament/ament_cmake.git
    version: humble
```
And here is the resulting output of vcs:
```
$ vcs import --input example.repos src
Input data is not valid format: 'NoneType' object is not subscriptable
```

christophebedard:
Fixes #255

The parsing code (`get_repos_in_vcstool_format()`) already provides a useful error message when an attribute is missing (e.g., `url:`). However, as shown in the issue above, it doesn't handle/detect empty repository entries. Therefore, check if the repository is empty before trying to access its attributes. Otherwise, provide a useful error message:

```
$ cat bad.repos
repositories:
  empty_entry:
  missing_url:
    type: git
$ vcs validate --input bad.repos 
Repository 'empty_entry' is empty
Repository 'missing_url' does not provide the necessary information: 'url'
```

Before this change, validation fails on `empty_entry` (and does not validate `missing_url`):

```
$ vcs validate --input bad.repos 
Input data is not valid format: 'NoneType' object is not subscriptable
```

I've also updated the tests to cover this.

### DIFF
--- a/test/bad.repos
+++ b/test/bad.repos
@@ -1,0 +1,4 @@
+repositories:
+  empty_entry:
+  missing_url:
+    type: git
--- a/test/test_commands.py
+++ b/test/test_commands.py
@@ -16,6 +16,7 @@ REPOS_FILE_URL = file_uri_scheme + REPOS_FILE
 REPOS2_FILE = os.path.join(os.path.dirname(__file__), 'list2.repos')
 TEST_WORKSPACE = os.path.join(
     os.path.dirname(os.path.dirname(__file__)), 'test_workspace')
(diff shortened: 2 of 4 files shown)


caioaamaral:
friendly ping @dirk-thomas

PR #257, opened 2023-03-26
Update CI config

christophebedard:
* Bump actions/checkout and actions/setup-python
* Remove Python 3.5 & 3.6, since those releases are no longer active
* Add 3.9 & 3.10
* Fix `test_flake8` to support newer versions of flake8
* Set `application-import-names` and update/sort imports
* Skip `test_flake8` on macOS & Windows, since _something_ seems to be different compared to Ubuntu, which leads to a lot of violations, but we don't really need to test that on all platforms


christophebedard:
Oh, I just realized there's already an open PR for this: https://github.com/dirk-thomas/vcstool/pull/247

PR #267, opened 2024-01-19
Support git insteadOf config entries during import

smith-doug:
ls-remote --get-url will apply any special insteadOf rules.  This prevents false "Path already exists and contains a different repository" errors if you have something like the below in your gitconfig. 
```
[url "git@github.com:"]
    insteadOf = https://github.com/
```

### DIFF
--- a/test/list_ssh.repos
+++ b/test/list_ssh.repos
@@ -1,0 +1,5 @@
+repositories:  
+  vcstool:
+    type: git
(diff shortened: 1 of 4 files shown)


christophebedard:
Would probably be good and easy enough to add a test here: https://github.com/dirk-thomas/vcstool/blob/master/test/test_commands.py

smith-doug:
Ty for reminding me about tests.  The extra output I added broke them.  Will update.

Issue #198, opened 2021-01-04
vcs export not detecting nested submodules


tokr-bit:
# Brief Description
If a repository contains multiple `git submodule`s, these submodules are not detected by calling `vcs export --nested`

# Steps to reproduce

```console
$ cd /tmp
$ mkdir ws_with_submodules && cd ws_with_submodules && git submodule add https://github.com/dirk-thomas/vcstool.git
$ vcs export --nested
```

**Output:**
 
```console
repositories
ws_with_submodules: Could not determine ref: fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
HEAD
```

**expected output**:

As `vcstool` is a `git repository`, the repository should be detected.

# Additional information

`tree -L 2 /tmp/ws_with_submodules` yields

```console
ws_with_submodules/
└── vcstool
    ├── CONTRIBUTING.md
    ├── LICENSE
    ├── Makefile
    ├── MANIFEST.in
    ├── publish-python.yaml
    ├── README.rst
    ├── scripts
    ├── setup.py
    ├── setup.sh
    ├── stdeb.cfg
    ├── test
    ├── vcstool
    └── vcstool-completion

5 directories, 9 files
```

dirk-thomas:
> `mkdir ws_with_submodules && cd ws_with_submodules && git submodule add https://github.com/dirk-thomas/vcstool.git`

When running the provided step I only get the following error: `fatal: not a git repository (or any of the parent directories): .git`

@tokr-bit Can you please post a [SSCCE](http://sscce.org/). Please describe what your expected behavior is. Thanks.

tokr-bit:
I forgot to add `git init`. I.e.:

```
$ cd /tmp
$ mkdir ws_with_submodules && cd ws_with_submodules 
$ git init && git submodule add https://github.com/dirk-thomas/vcstool.git
$ vcs export --nested
```

produces the error mentioned above. The expected behavior remains the same as mentioned above.

dirk-thomas:
The error message you are seeing has nothing to do with nesting. The just created empty repository in `/tmp/ws_with_submodules` has no commits yet. Please try it again with a non-empty repository.

dirk-thomas:
@tokr-bit Friendly ping.

tokr-bit:
```
$ cd /tmp
[...]

PR #272, opened 2024-04-07
Add git submodule support (to detect submodule repos)

varunsairam:
Found that git submodules were not getting detected by `vcs`. Believe the issue comes down to file structure and type. 

(Also faced in issue: #198) 

### Typical git repo
```
git_repo
    |--- .git (directory)
```

### Submodule repo
```
outer_repo
    |--- .git (directory)
    |--- submodule_repo
         |--- .git (file)
```

Contents of `.git` file (the one inside the submodule_repo): 
```
gitdir: ../.git/modules/submodule_repo
```

So effectively, submodules don't have `.git` directory of their own but rather a `.git` linked file that links to the directory in the outer `.git` directory.

Currently, it appears that vcstool is designed to determine if a repo is a git repo by checking for only a `.git` directory:
https://github.com/dirk-thomas/vcstool/blob/daf389377310f7f31b2171f51a79af82c31bf3e2/vcstool/clients/git.py#L31

Tried to fix this here. The submodules now seem to get detected as git repos. Hope there would be no unintended consequences though?

### DIFF
--- a/vcstool/clients/git.py
+++ b/vcstool/clients/git.py
@@ -28,7 +28,8 @@ class GitClient(VcsClientBase):
 
     @staticmethod
     def is_repository(path):
-        return os.path.isdir(os.path.join(path, '.git'))
+        return (os.path.isdir(os.path.join(path, '.git')) or 
+                os.path.isfile(os.path.join(path, '.git')))
 
     def __init__(self, path):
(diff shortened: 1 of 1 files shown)


rty813:
really need this feature!!!

PR #283, opened 2025-02-05
feat: add partial clone argument in vcs import

Owen-Liuyuxuan:
### Introduction

A partial clone is an efficient way to clone only part of the git history of a certain repository. 

For example, we can clone big repositories with an extensive Git history using partial clones and save much time and networks.
```
git clone --filter=blob:none --no-checkout https://github.com/autowarefoundation/autoware.universe
git checkout $tag # This could be either a branch or a random tag in the history
```
At the time the PR was created, the partial clone could shrink the download size of `autoware.universe` from 1.5G (1.1G .git + 340M codes) to 500M (160M.git + 340M codes).

### Implementation

In this PR, I incorporate the argument in the `vcs import` command for Git client. The codes are straight-forward, but it solve many of the problems we faced.


andry81:
Why not to always filter out the rest of the history on clone or fetch?
The tool basically is used for an external repository checkout into a directory of the root repository. And most of the time an external repository does not require locally the history content. So the flag better to use by default and add a yaml key to force use the full history checkout.

Owen-Liuyuxuan:
@andry81 
From our experience, we also want to set that by default, but I believe I should respect the default behavior of the code for now.

andry81:
@Owen-Liuyuxuan 
Last PR were merged 3 years ago, so you may want to read this: https://github.com/dirk-thomas/vcstool/issues/242

Owen-Liuyuxuan:
@andry81 
Thank you for the information, it is also fine to put this PR as an open reference.

PR #291, opened 2025-09-18
parse shell key in repo yaml and execute the specified commands

michael-elomar:
The idea is to write small simple shell commands in the yaml manifest under each repo, to be executed after the repo is cloned during the import command when initializing a repository. It is an optional field int the yaml manifest, it doesn't have to be included, but I personally found it helpful for me, like for symlinking files inside a certain repo to the global workspace directory for example.

I would love to hear your thoughts.

### DIFF
--- a/vcstool/clients/vcs_base.py
+++ b/vcstool/clients/vcs_base.py
@@ -6,7 +6,7 @@ from urllib.error import HTTPError
 from urllib.error import URLError
 from urllib.request import Request
 from urllib.request import urlopen
-
+import shlex
 
 class VcsClientBase(object):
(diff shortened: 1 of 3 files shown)


andry81:
Recently a repository fork tree has been broken:
https://github.com/useful-forks/useful-forks.github.io/issues/76  
https://github.com/orgs/community/discussions/173970  

But the latest *most active* fork is: https://github.com/ros-infrastructure/vcs2l

May be send the PR there?

michael-elomar:
Alright, I'll try doing that. Thank you.

